PR #4, opened 2023-11-27
Glm/fragment aggregator

glehmannmiotto:
This PR introduces the Fragment Aggregator for the readout application.
It also makes use of the modified NetworkConnection class in the coredal (feature/service_exposure branch) schema.
More testing is needed on the connectivity establishment in the generate_modules code, but at least things compile and seem correct at first sight.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,7 @@ find_package(coredal REQUIRED)
 
 #find_package(Boost COMPONENTS unit_test_framework REQUIRED)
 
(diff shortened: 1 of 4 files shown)


gcrone:
I think this looks OK. It took me a while to update my test database to satisfy all the connections required to run `gen_readout_modules` and I added a few more checks of the form
```
if (dlhInputQDesc == nullptr) {
  throw (BadConf(ERS_HERE,
}
if (dlhReqInputQDesc == nullptr) {
   throw (BadConf
```
to `ReadoutApplication::generate_modules` after seeing it segfault.

PR #6, opened 2023-12-06
Reorganisation of appdal schema

gcrone:
In this PR, the main part of the application schema and the FD specific parts are split into separate schema files. This is to facilitate a possible future split into common, FD and ND repositories.

The main part is renamed from `readout.schema.xml` to  `application.schema.xml` to better reflect its contents.

 Also, the `gen_readout_modules` test program is renamed  to the more generic generate_modules_test


glehmannmiotto (approved):
Looks good to me

PR #7, opened 2023-12-06
restructure readout modules schema to follow closer daq module code o…

glehmannmiotto:
…rganization.

Work is not complete as the views files need to be updated and the diagrams recreated for the documentation.
I also need advice from Roland on which of the parameters belong to any request handler and which ones are specific to only some request handler configurations (and same for latency buffers).

### DIFF
--- a/schema/appdal/application.schema.xml
+++ b/schema/appdal/application.schema.xml
@@ -233,18 +233,6 @@
   <attribute name="preallocation" type="bool" init-value="true"/>
  </class>
 
- <class name="FixedRateLatencyBuffer">
-  <superclass name="LatencyBuffer">
- </class>
-
- <class name="BinarySearchLatencyBuffer">
-  <superclass name="LatencyBuffer">
- </class>
-
- <class name="SkipListLatencyBuffer">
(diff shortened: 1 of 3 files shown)


schema/appdal/application.schema.xml
```diff
@@ -222,16 +233,33 @@
...
+
+ <class name="BinarySearchLatencyBuffer">
+  <superclass name="LatencyBuffer">
+ </class>
+
+ <class name="SkipListLatencyBuffer">
+  <superclass name="LatencyBuffer">
+ </class>
+
+ <class name="ReadoutModelConf">
```
gcrone, line 236: ReadoutModelConf or ReadoutModuleConf?

gcrone (approved):
Looks good. I knew there were some things that didn't quite match between the drawio diagram from the summer and the actual module code!

gcrone:
And of course I forgot to fix the network connection descriptor after changing coredal!!

schema/appdal/application.schema.xml
glehmannmiotto: the class is Model, not Module in the c++ readout code.

PR #15, opened 2024-01-12
Eflumerf/df application

eflumerf:


glehmannmiotto (commented):
- operational_environment is also set as attribute in the DetectorConfig class (op_env) related to Session. Are those independent?
- Looking at the example configuration there is one thing that is not nice in the definition of the DataStoreConf class. It is in principle technology agnostic, but then it has a relation specifically to HDF5FileLayoutParams. If we want to support the possibility of using different storage technologies we should have an abstract base class like "LayoutParams" and different implementations of it  would determine the storage technology used (making also the "type" attribute superfluous).

eflumerf:
Your first point is why I made https://github.com/DUNE-DAQ/dfmodules/commit/3c3c4132e8370393f592a83dba67c817a3f682fd, so that the DataWriter and TPStreamWriter get Session/DetectorConfig/op_env instead of having a separate configuration parameter "operational_environment".

Probably, I'd call it "FileParams", but yes, I think that is a good idea.

glehmannmiotto:
how is the hwconfig related to the location where we store data files?

gcrone (approved):
All looks good

PR #18, opened 2024-01-18
Initial trigger OKS schema implementation

ArturSztuc:
Rather large PR, with the first trigger OKS implementation. It should work together with the [feature/OKS](https://github.com/DUNE-DAQ/trigger/tree/feature/OKS) trigger branch.
Implementations of modules that should work:
* `ModuleLevelTrigger`
* `CustomTriggerCandidateMaker` (fakes given algorithm, creating an empty TC of configured type)
* `TimingTriggerCandidateMaker` (for HSIEvents)
* `RandomTriggerCandidateMaker` (makes TCs and spaces them randomly in time)

Things that affect other people's code:
* I replaced the code repeated for each application inside `pybindsrc/dal_methods.cpp` with a templated version. Please check that it makes sense!
* Added `HSIEvents` service inside `test/config/test-session.data.xml`, but I don't know who else should look into this.
* There are of the changes in `schema/appdal/application.schema.xml` and `src/ReadoutApplication.cpp` that I don't remember making -- I think they are because I was initially using `git rebase` rather than `git merge` for catching up with the develop branch...

Other notes:
* At the moment this is supposed to be a simple application implementation, without any tees, buffers etc. So a `TriggerCandidate` makers push their TCs into the queue that goes directly into `MLT`, without putting them into the buffer on the way.
* A lot of the commented-out code in the `TriggerApplication.cpp` is to make all the tees, buffers etc. Buffering still doesn't 100% work, and it will all probably be replaced by a completely different structure, more similar to how modules are structured in the Readout.
* At the moment the MLT cannot do Region Of Interest grouping/readout with this OKS change.


schema/appdal/application.schema.xml
```diff
@@ -331,8 +331,8 @@
 
  <class name="SmartDaqApplication" is-abstract="yes">
   <superclass name="Application"/>
-  <relationship name="queue_rules" class-type="QueueConnectionRule" low-cc="zero" high-cc="many" is-composite="no" is-exclusive="no" is-dependent="no"/>
-  <relationship name="network_rules" class-type="NetworkConnectionRule" low-cc="zero" high-cc="many" is-composite="no" is-exclusive="no" is-dependent="no"/>
+  <relationship name="queue_rules" class-type="QueueConnectionRule" low-cc="one" high-cc="many" is-composite="no" is-exclusive="no" is-dependent="no"/>
```
gcrone: I think it is quite conceivable that a SmartDaqApplication has only one DAQModule and therefore no Queues so why force low-cc to one rather than zero?

ArturSztuc: I have to admit that I don't remember making that change, and assumed it's because of using git rebase rather than git merge in the past. Definitely agree with you here. Should I change it back to zero together with the `network_rules` below, or only for the `queue_rules`?

gcrone: Yes, ahh, I see you've done it already!

gcrone:
Why are there 3 specific candidate maker `relationships` in the TriggerApplication?
Wouldn't a single candidate_maker_conf `relationship` with a zero to many multiplicity be better. Making a CandidateMaker base class with a relationship to a CandidateMakerConf and having the 3 implementations inherit appropriately?

Also, what is the relationship between the Custom,Timing and Random CandidateMakers and the TriggerCandidateMaker associated with the TriggerAlgorithms?

ArturSztuc:
@gcrone  This is because these three classes in the are not actually related to each other in the trigger in any way, even though some of the internal logic is similar. The only related inheritance between them is that they are all `DAQModule`s. The configs are completely different though, and there is no inheritance/relationship between them (or between them and normal `TriggerAlgorithm`s or `TriggerCandidate`s) in the trigger code itself.

This is maybe something to be changed in the future, but would require new development/extra features in the trigger, so for now it didn't make sense to me to include this in the schema. I had a chat with other trigger people exactly about this earlier today though, and maybe in the future we will do something smarter, maybe even with some inheritance from the generic trigger maker (and then the schema oks code would change in a way that you expected it too).

gcrone:
The fact that the DAQModule candidate maker classes have no inheritance doesn't mean that the schema classes can't have inheritance. The code in TriggerApplication::generate_modules() looks like it is doing the same thing for all 3. A CandidateMakerConf class with a `template_for` attribute specifying which DaqModule class it goes with could be the parent of all 3 xxxCandidateMakerConf classes and they need share nothing else. Similarly a CandidateMaker class which inherits from DaqModule with a relationship to CandidateMakerConf could be used as the parent for all 3 modules. It has no implications for the implementation of the actual DAQModule plugins.

ArturSztuc:
I was going to make a templated function (similar to what I've done in [pybindsrc/dal_methods.cpp](https://github.com/DUNE-DAQ/appdal/pull/18/files#diff-dd3bdb5a43a63429b14eeed34e8f8772ef77e081e7e5752f324873978d32e557)), but new structure with `template for` should work just as well for this... I'll try to work on it now, but it might take a while.
I guess this is also a step towards moving trigger structure to be more similar to readout (more about it in
https://indico.fnal.gov/event/62280/), which is what I was going to work on next with OKS.

src/TriggerApplication.cpp
```diff
@@ -0,0 +1,509 @@
...
+   * Get all the queue connection descriptions
+   **************************************************************/
+  for (auto rule : get_queue_rules()) {
+    std::string destination_class = rule->get_destination_class();
+    std::string data_type = rule->get_descriptor()->get_data_type();
+    if((destination_class == "ModuleLevelTrigger") && (data_type == "TriggerCandidate"))
+      tcQueueDesc = rule->get_descriptor();
+  }
+  if(!tcQueueDesc)
+    throw (BadConf(ERS_HERE, "No description for the queue of TCs going tinto the MLT given"));
```
gcrone, line 131: Should really have braces around the if action

src/TriggerApplication.cpp
```diff
@@ -0,0 +1,509 @@
...
+  if(!tdMLTNetDesc)
+    throw (BadConf(ERS_HERE, "No MLT network connection for the output TriggerDecision given"));
+
+  /**************************************************************
+   * Get the MLT
+   **************************************************************/
+  auto mlt_conf = get_mlt_conf();
+  // Don't build the rest if we have no MLT
+  if(!mlt_conf)
+    return modules;
```
gcrone: If you don't have an MLT config you return an empty module list? Isn't this an invalid configuration that should throw a BadConf Issue?

src/TriggerApplication.cpp
```diff
@@ -0,0 +1,509 @@
...
+
+  // Get the triggering algos handler
+  auto trigger_algs = get_trigger_algs();
+  // Don't build the rest if we don't have triggering algs
+  if(trigger_algs.size() == 0)
+    return modules;
+
+  // Make the TPChannelFilters
+  auto tpChannelFilterConf = get_tpchannelfilter_conf();
```
gcrone, line 373: Not sure what you're doing with the TPChannelFilterConf, I suppose this is still a work in progress

ArturSztuc: Yes, tpChannelFilter is very much work in progress, and will be probably completely changed in the future

ArturSztuc:
@gcrone The most recent commit ([23be1ab](https://github.com/DUNE-DAQ/appdal/pull/18/commits/23be1ab422564f6462d47b6b2bdfc140e725796e)) does what you recommended, please let me know what you think.

gcrone (approved):
I think that looks OK now. A couple of suggestions though, you could add a `template_for` attribute to the individual xxStandaloneCandidateMakerConf classes with the correct init-value. That will override the default in the base class. Also the StandaloneCandidateMakerConf class itself should probably have  is-abstract="yes".

ArturSztuc:
Thank you for the suggestions, both implemented & committed in [d63a814](https://github.com/DUNE-DAQ/appdal/pull/18/commits/d63a81486dd4b0c32c0e526613188f200be7a5ab)

PR #19, opened 2024-01-19
Glm/restruct

glehmannmiotto:
Restructure the organisation of data files so they are easy to reuse.


gcrone (approved):
All looks good to me. I had already decided it would make sense to move the connections to a separate file. Moving some of the files from test/config into config/appdal is a good idea for reusing in other places. Although I wonder whether just config without the appdal would be better since it would make it easier to move e.g. fsm.data.xml to another more general place later (logically fsm maybe belongs to coredal).

eflumerf:
How do we want to sequence this with https://github.com/DUNE-DAQ/appdal/pull/18? It looks like there will have to be some merges done...I think it may be better to merge the trigger stuff first, then update this branch (moving the trigger module config and the extra connection definition appropriately)

PR #24, opened 2024-02-07
Eflumerf/multiple streams

eflumerf:


test/config/df-segment.data.xml
```diff
@@ -99,14 +99,51 @@
  <rel name="data_writer" class="DataWriterConf" id="dw-01"/>
  <rel name="uses" class="DFHWConf" id="dfhw-01"/>
 </obj>
+	
+	<obj class="DFApplication" id="df-02">
```
gcrone, line 103: Why are these lines indented differently to the other objects?

test/config/trigger-segment.data.xml
```diff
@@ -116,7 +116,9 @@
  <attr name="application_name" type="string" val="daq_application"/>
  <rel name="runs_on" class="VirtualHost" id="vlocalhost"/>
  <rel name="contains">
-  <ref class="ReadoutGroup" id="dummyRog"/>
+  <ref class="ReadoutGroup" id="rog-1"/>
+	 <ref class="ReadoutGroup" id="rog-2"/>
+	 <ref class="ReadoutGroup" id="rog-3"/>
```
ArturSztuc, line 121: Something happened here with indentation

schema/appdal/trigger.schema.xml
ArturSztuc, line 1: There's a lot of changes here that will take me a while to review. Is this all only just rearranging, or are there new additions/deletions here too? If it's the latter, then I will try to run this too.

gcrone (approved):
Looks OK to me

ArturSztuc (approved):
LGTM

test/config/df-segment.data.xml
eflumerf: Likely due to a different editor. Let me do a dbe_main pass to clean it up.

PR #26, opened 2024-02-09
Eflumerf/fake hsi app

eflumerf:


glehmannmiotto (approved):
Approving, to get the schema in for hsilibs, but the configuration is not correct, I think. You can merge and tomorrow we can fix the example conf.

PR #27, opened 2024-02-17
schema and algorithm adjustments for revised trigger structure

glehmannmiotto:
This branch can only be merged after:
cordal gcrone/jsonable

It should be merged together with the glm/newtrig branches of:
appdal
readoutlibs
fdreadoutlibs
fdreadoutmodules
trigger

Testing is still to be completed. I need help to converge on this.


test/config/ru-segment.data.xml
```diff
@@ -169,6 +169,6 @@
   <ref class="ReadoutApplication" id="ru-03"/>
  </rel>
  <rel name="controller" class="RCApplication" id="ru-controller"/>
-</obj>
+ </rel></obj>
```
gcrone: There should not be a `</rel>` here!

config/appdal/moduleconfs.data.xml
gcrone, line 1: Not really this file, but  there is a reference to the `TimingTriggerCandidateMaker` class which no longer exists in config/appdal/connections.data.xml.

glehmannmiotto:
I fixed a few things in the configurations and created the MLTApplication which should, in principle work like before. To be tested...

schema/appdal/trigger.schema.xml
```diff
@@ -99,19 +99,52 @@
...
+  <relationship name="signals" class-type="HSISignalWindow" low-cc="one" high-cc="many" is-composite="no" is-exclusive="no" is-dependent="no"/>
+ </class>
+
+  <class name="HSISignalWindow">
+  <attribute name="signal_type" type="u32" init-value="0" is-not-null="yes"/>
+  <attribute name="time_before" type="u32" init-value="10000" is-not-null="yes"/>
+  <attribute name="time_after" type="u32" init-value="20000" is-not-null="yes"/>
+ </class>
+
+ <!--
```
gcrone: The genconfig code generation still tries to build this commented out class and fails. Maybe we need to update `genconfig` but the schema editors will remove the comment if anybody edits the file anyway so it's best not to rely on comments.

src/MLTApplication.cpp
```diff
@@ -0,0 +1,284 @@
...
+    create_mlt_network_connection(tiMLTNetDesc->get_uid_base(), tiMLTNetDesc, confdb, dbfile);
+
+  oksdbinterfaces::ConfigObject tcMLTNetObj =
+    create_mlt_network_connection(tcMLTNetDesc->get_uid_base(), tcMLTNetDesc, confdb, dbfile);
+
+        // Network connection for the MLT: output TriggerDecision
+  oksdbinterfaces::ConfigObject tdMLTNetObj =
+    create_mlt_network_connection(tdMLTNetDesc->get_uid_base(), tdMLTNetDesc, confdb, dbfile);
+
+  oksdbinterfaces::ConfigObject* timesyncNetObj;
```
gcrone: Should be initialised to nullptr here!

schema/appdal/trigger.schema.xml
gcrone: Actually, I just checked, it's not genconfig itself that ignores the comment delimiters, it's the `daq_oks_codegen` function in daq-cmake getting the list of classes through a grep.

config/appdal/connections.data.xml
```diff
@@ -105,6 +110,20 @@
  <rel name="associated_service" class="Service" id="HSIEvents"/>
 </obj>
 
+<obj class="NetworkConnectionDescriptor" id="ta-publisher">
+ <attr name="uid_base" type="string" val="ta_"/>
+ <attr name="connection_type" type="enum" val="kSendRecv"/>
```
ArturSztuc: I remember you mentioned that the TA will be published, meaning if we have multiple algorithms types, we will need to filter TAs fron unwanted algorithm at TCMaker level (since there will be one TCMaker per algorithm type). 
Here though I see `kSendRecv` rather than `kPubSub`, does that mean that there can be only one receiver for that publisher? I know how `kPubSub` works (I think), but not so much for `kSendRecv`.

config/appdal/moduleconfs.data.xml
```diff
@@ -148,34 +181,48 @@
  <attr name="warn_on_empty_buffer" type="bool" val="1"/>
 </obj>
 
+<obj class="RequestHandler" id="def-tp-request-handler">
+ <attr name="handler_threads" type="u16" val="1"/>
+ <attr name="request_timeout" type="u32" val="10000"/>
+ <attr name="pop_limit_pct" type="float" val="0.5"/>
+ <attr name="pop_size_pct" type="float" val="0.8"/>
```
ArturSztuc, line 190: Does this mean 80% of the buffer is cleared when it's full? I always thought it would be a smaller number, like 20% or so (never checked it though). I wouldn't be surprised if removing 80% of it could cause issues, if latencies are long or/and rates are high.

src/MLTApplication.cpp
ArturSztuc, line 1: So the standalone TCMakers (`CustomTriggerCandidateMaker`, `RandomTriggerCandidateMaker`) still are dealt in the same way, with an exception of the `TimingTriggerCanidateMaker`, which is moved to the new architecture via `HSI2TCTranslatorConf`?

test/config/trigger-segment.data.xml
```diff
@@ -117,56 +137,84 @@
...
-  <ref class="NetworkConnectionRule" id="hsi-rule"/>
+  <ref class="NetworkConnectionRule" id="ta-net-rule"/>
+  <ref class="NetworkConnectionRule" id="tp-net-rule"/>
+  <ref class="NetworkConnectionRule" id="data-req-trig-net-rule"/>
  </rel>
- <rel name="contains">
-  <ref class="ReadoutGroup" id="rog-1"/>
-  <ref class="ReadoutGroup" id="rog-2"/>
-  <ref class="ReadoutGroup" id="rog-3"/>
+ <rel name="data_subscriber" class="DataReaderConf" id="ta-subscriber-1"/>
```
ArturSztuc, line 196: Should this be some form of tp-subscriber, given it's ta-maker module?

ArturSztuc (approved):

glehmannmiotto:
I fixed most comments. For now the communication between modules for the trigger is send/recv, while we are debugging. Then we can switch back to pub/sub.

PR #28, opened 2024-02-28
Add Hermes controller related classes

gcrone:


schema/appdal/fdmodules.schema.xml
```diff
@@ -132,6 +132,28 @@
...
  </class>
 
+ <class name="HermesController">
+  <superclass name="DaqModule"/>
+  <superclass name="IpbusDevice"/>
+  <attribute name="port" type="u32" init-value="17476" is-not-null="yes"/>
+  <relationship name="links" class-type="HermesLinkConf" low-cc="one" high-cc="many" is-composite="no" is-exclusive="no" is-dependent="no"/>
+ </class>
+
+ <class name="HermesLinkConf">
```
glehmannmiotto, line 142: In a future iteration we should think how to avoid the 2 relations to NICReceiver and DROStreamConf which may cause inconsistent configurations. But for now we can live with this.

glehmannmiotto (approved):
ok to proceed.

PR #33, opened 2024-03-07
Enabling the TC-making application & increasing prescale

ArturSztuc:
To merge after https://github.com/DUNE-DAQ/trigger/pull/280 is merged.

* Enabled tc-making application, since the PR above makes it work.
* Increased TC-making prescale, I was getting a bit high rates with TC prescale of 2.
* Removed the `"TIMING_SHARE": "getenv",` line from `boot.json`, it makes nanorc crash...

### DIFF
--- a/test/boot.json
+++ b/test/boot.json
@@ -15,10 +15,15 @@
       "host": "local",
       "port": 3339
     },
-    "mlt": {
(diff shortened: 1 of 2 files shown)


gcrone (approved):
Looks OK to me

PR #35, opened 2024-03-08
Replace dromap_CRP4 with freshly generated version from dromap2oks

gcrone:
Old version had spurious `HermesLinkConf` objects with ids>1 and also since PR33 `id` should be `link_id`


glehmannmiotto (approved):
LGTM

PR #36, opened 2024-03-12
Asztuc/adding tcbuffering

ArturSztuc:
Adding TC buffering. It works, and TC fragments from the trigger's latency buffers are successfully saved in the hdf5 files!
I had to make few small changes to accomodate for that, and a small re-write of how MLT sits in OKS might be needed in the future. Issues:

* TC maker module subscribes to network publishing TriggerActivities, with `endpoint_class` of `DataSubscriber`. For buffering, it needs to publish `TriggerCandidates` with `endpoint_class` of `DataSubscriber`, and by default `DataSubscriber` endpoint was treated as inputs-only in `TriggerApplication.cpp`.
* TC buffer module needs input connection (for TCs to go in), and output network connection (seems to be enforced by `ReadoutModule` structure, even if the module is not actually publishing anything).
* With both of these in mind, I have two network connections for TCs. One with `endpoint_class` of `DataSubscriber` -- that's what's effectively being used everywhere, and one with `endpoint_class` of `ModuleLevelTrigger` -- a dummy, the buffering module needs output network to work, despite not publishing anything.

I think long-term the TC buffering should be moved to the module handling MLT, which means moving MLT to the ReadoutModule-like structure too. I think this would make sense, and will probably try to work on this next.

I hope someone has better/simpler ideas from what I've done for this PR though...

### DIFF
--- a/config/appdal/connections.data.xml
+++ b/config/appdal/connections.data.xml
@@ -125,6 +125,13 @@
  <rel name="associated_service" class="Service" id="triggerCandidates"/>
 </obj>
 
+<obj class="NetworkConnectionDescriptor" id="tc-mlt-desc">
+ <attr name="uid_base" type="string" val="tcs_mlt_"/>
+ <attr name="connection_type" type="enum" val="kPubSub"/>
+ <attr name="data_type" type="string" val="TriggerCandidate"/>
+ <rel name="associated_service" class="Service" id="triggerCandidates"/>
+</obj>
+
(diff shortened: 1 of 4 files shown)


glehmannmiotto (approved):
This looks good, considering the limitation of having to set a network output connection (which we should fix in the readout model).
I agree with you that it would be simpler to move the MLT to the same structure as the other apps, and this getting the TC buffering "for free".

gcrone (approved):
Looks OK

PR #38, opened 2024-03-12
Make a SourceIDConf an optional relationship in SmartDaqApplication, to

eflumerf:
make it easier for the TRB to build maps for DataRequests (could also be useful in MLTApplication.cpp). Update configurations.


glehmannmiotto (commented):
You changed the dal algorithms in a way that preserves the code in other packages, which is nice.
The only thing I don't like is that ReadoutApplication extends from SmartApplication, but, instead of using the new relation it has two numeric values for the TP and TA source IDs.

glehmannmiotto (approved):
LGTM

gcrone:
Trying to run the test-session with this branch (with develop and drunc branches merged), the tp-stream-writer throws a BadConf issue with 'No SourceIDConf given to TPWriterApplication!'. Should the commented out 'Get TPSet sources' code be uncommented or should something be added to the configuration DB?

PR #41, opened 2024-03-14
Drunc integration

plasorak:
Add command line parameters, env variable and rte script to be used by drunc's process manager and controllers.


plasorak:
This needs to be used in conjunction with https://github.com/DUNE-DAQ/coredal/pull/16

glehmannmiotto (approved):
OK for this release, but we should construct the rest URI for commands from the "runs_on" relationship information in the future.

test/config/df-segment.data.xml
```diff
@@ -87,6 +87,16 @@
...
  <attr name="application_name" type="string" val="daq_application"/>
+ <attr name="commandline_parameters" type="string">
+  <data val="--name"/>
+  <data val="df-01"/>
+  <data val="-c"/>
+  <data val="rest://localhost:3339"/><!-- ALTER_ME -->
+  <data val="-i"/>
+  <data val="kafka://monkafka.cern.ch:30092/opmon"/>
+  <data val="--configurationService"/>
+  <data val="oksconfig:test/config/test-session.data.xml"/><!-- ALTER_ME -->
```
gcrone, line 98: I know this is a hack to get things working but these these thing really do not belong here.  This not only adds many duplicated lines throughout the configuration but it also makes the Segment usable i only one Session. Why does drunc not construct the command line from OKS variables and its own command line?

glehmannmiotto:
Also, as a note for the future, we should add the specific environment variables needed for apps (e.g. ERS streams)

test/config/df-segment.data.xml
glehmannmiotto: that would be the better option. For the future.

plasorak: The rationale for doing this was:
 - I don't want to modify the configuration on the fly in drunc.
 - I'm unsure if there are "placeholders" in OKS or if I should set some up myself.
I think the way to do this without adding too much complexity or changing the appfwk CLI is to add environment variables and set them up in the session (which I probably should have thought of earlier). This likely also touches on things discussed in the appfwk review.

I agree this isn't ideal.

For the command facility, it's a different problem: if it isn't set up correctly, then the parent controller doesn't know how to talk to their children (and therefore has to parse the CLI somewhere in drunc, which is not nice).

PR #42, opened 2024-03-19
Gcrone/diagrams

gcrone:
Basically just updating the schema diagrams but while updating the fd_cusomizations view I noticed that the default `eal_args` string in NICReceiverConf was not the empty string that we need to run the CRP4 config.


glehmannmiotto (approved):
LGTM

PR #44, opened 2024-03-20
Retrieve SourceIDConf for the DF app and use it to configure the TRB.…

eflumerf:
… Change module UIDs to reflect app first then module type, and start laying the groundwork for if we want to support multiple datawriters in the future


gcrone (approved):
LGTM

PR #45, opened 2024-03-21
DAQ App CLI env vars

plasorak:
This PR:
- Uses environment variables in the CLI of the applications, such that we don't need to manually specify things that get reused, and can change them once and for all in the Session DAL.
- Correct a couple of bugs in the name of the application CLI.
- Makes the HSI work with drunc.

### DIFF
--- a/test/config/hsi-segment.data.xml
+++ b/test/config/hsi-segment.data.xml
@@ -90,7 +90,7 @@
   <data val="--name"/>
   <data val="hsi-01"/>
   <data val="-c"/>
-  <data val="rest://np04-srv-019:3382"/><!-- ALTER_ME -->
+  <data val="rest://localhost:3382"/><!-- ALTER_ME -->
   <data val="-i"/>
   <data val="${DAQAPP_CLI_INFO_SVC}"/>
   <data val="--configurationService"/>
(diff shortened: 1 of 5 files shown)


test/config/hsi-segment.data.xml
```diff
@@ -86,6 +86,16 @@
  <attr name="application_name" type="string" val="daq_application"/>
  <rel name="runs_on" class="VirtualHost" id="vlocalhost"/>
  <rel name="source_id" class="SourceIDConf" id="hsi-srcid-01"/>
+ <attr name="commandline_parameters" type="string">
+  <data val="--name"/>
+  <data val="hsi-01"/>
+  <data val="-c"/>
+  <data val="rest://np04-srv-019:3382"/><!-- ALTER_ME -->
```
gcrone: np04 host name? Everything in the test config was set up to use localhost so anybody can run it anywhere!
I just greppped for 'rest:' and I see there are some np04 references in df-segment as well

plasorak: Woops, yes that's a mistake

gcrone (approved):
Looks OK now

PR #47, opened 2024-04-03
Glm/mlt

glehmannmiotto:
MLT restructuring according to presentation ad DS/PP meeting

### DIFF
--- a/config/appdal/moduleconfs.data.xml
+++ b/config/appdal/moduleconfs.data.xml
@@ -152,18 +152,9 @@
 </obj>
 
 <obj class="ModuleLevelTriggerConf" id="def-mlt-conf">
- <attr name="hsi_trigger_type_passthrough" type="bool" val="0"/>
- <attr name="merge_overlapping_tcs" type="bool" val="1"/>
  <attr name="td_out_of_timeout" type="bool" val="1"/>
  <attr name="buffer_timeout" type="u32" val="1"/>
  <attr name="td_readout_limit" type="u32" val="1"/>
- <attr name="use_bitwords" type="bool" val="0"/>
- <attr name="use_roi_readout" type="bool" val="0"/>
- <attr name="use_readout_map" type="bool" val="0"/>
- <rel name="tc_readout_map">
-  <ref class="TCReadoutMap" id="def-tc-map"/>
-  <ref class="TCReadoutMap" id="def-hsi-tc-map"/>
- </rel>
 </obj>
 
 <obj class="NICReceiverConf" id="def-emu-nic-receiver-conf">
--- a/config/appdal/moduleconfs.data.xml
+++ b/config/appdal/moduleconfs.data.xml
@@ -104,6 +104,17 @@
  <attr name="free_threshold" type="s32" val="1"/>
 </obj>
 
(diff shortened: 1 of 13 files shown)


config/appdal/moduleconfs.data.xml
```diff
@@ -117,6 +128,20 @@
...
+  <ref class="HSISignalWindow" id="def-hsi-signal-window"/>
+ </rel>
+</obj>
+
+<obj class="HSISignalWindow" id="def-hsi-signal-window">
+ <attr name="signal_type" type="u32" val="1"/>
+ <attr name="time_before" type="u32" val="10000"/>
+ <attr name="time_after" type="u32" val="20000"/>
+</obj>
+
```
ArturSztuc, line 144: Not for this PR, but we will need to think about HSI a little more in the future for v5.
There were large changes to how we deal with them in v4 because of the new CTB, CIB and PNS signals that are also going through HSIEvents but to separate modules, and even the standard `TimingTriggerCandidateMaker` configuration was changed substantially for some of that.

schema/appdal/trigger.schema.xml
```diff
@@ -106,18 +106,17 @@
   <relationship name="configuration" class-type="DataReaderConf" low-cc="one" high-cc="one" is-composite="no" is-exclusive="no" is-dependent="no"/>
  </class>
 
+ <class name="HSI2TCTranslatorConf">
+  <superclass name="DataReaderConf"/>
+  <relationship name="signals" class-type="HSISignalWindow" low-cc="one" high-cc="many" is-composite="no" is-exclusive="no" is-dependent="no"/>
+ </class>
+
```
ArturSztuc, line 113: Ditto here for the dealing with HSI, but again, not for this PR.

schema/appdal/trigger.schema.xml
```diff
@@ -278,6 +283,8 @@
 
  <class name="TriggerDataHandler">
   <superclass name="ReadoutModule"/>
+  <relationship name="enabled_source_ids" description="List of source IDs enabled in this session: used by trigger to define the readout window for trigger decisions." class-type="SourceIDConf" low-cc="zero" high-cc="many" is-composite="no" is-exclusive="no" is-dependent="no"/>
+  <relationship name="mandatory_source_ids" description="Source Ids that will always be included in a trigger decision." class-type="SourceIDConf" low-cc="zero" high-cc="many" is-composite="no" is-exclusive="no" is-dependent="no"/>
```
ArturSztuc, line 287: Why both `"TCDataProcessor"` and `"TriggerDataHandler"` (and therefore all the TAMakers, TCMakers too) need this?

ArturSztuc (approved):
LGTM!

Unfortunately I didn't have the time to run the code, but did look through it.
If I understand it right, the TD-making and MLT are still in one application, but in separate DAQ modules? 

I left few comments in general, but some not for this PR, and definitely nothing to block the merge.

schema/appdal/trigger.schema.xml
glehmannmiotto: The TCDataProcessos has the mandatory links which are a static list (for the configuration). Out of this information the TriggerDataHandler is generated and when it's generated is also is associated to the list of enabled source IDs.

PR #49, opened 2024-04-09
Command line param method in Application DAL

plasorak:
This PR is related to https://github.com/DUNE-DAQ/coredal/pull/18

This PR contains:
 - an example of the implementation of the DAL changes described in `coredal` (new `OpmonService`, new control `Service` for each application)
 - an implementation of a configuration method to get the command line parameters for `SmartDaqApplication` (which uses the same templated function used for `DaqApplication` in `coredal/include/util.hpp`)
 - python bindings for the above method
 - some handy executable to access command line parameters and environment (python ones are in `scripts` and c++ in `app`)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,25 +16,25 @@ find_package(coredal REQUIRED)
 
 #find_package(Boost COMPONENTS unit_test_framework REQUIRED)
 
-daq_oks_codegen(application.schema.xml fdmodules.schema.xml trigger.schema.xml 
+daq_oks_codegen(application.schema.xml fdmodules.schema.xml trigger.schema.xml
   NAMESPACE dunedaq::appdal DEP_PKGS coredal)
 
+
 daq_add_library(ReadoutApplication.cpp SmartDaqApplication.cpp
-	DFApplication.cpp DFOApplication.cpp TPWriterApplication.cpp FakeDataApplication.cpp FakeHSIApplication.cpp TriggerApplication.cpp MLTApplication.cpp HSIEventToTCApplication.cpp 
+	DFApplication.cpp DFOApplication.cpp TPWriterApplication.cpp FakeDataApplication.cpp FakeHSIApplication.cpp TriggerApplication.cpp MLTApplication.cpp HSIEventToTCApplication.cpp
  LINK_LIBRARIES oksdbinterfaces::oksdbinterfaces okssystem::okssystem
   logging::logging coredal::coredal oks::oks ers::ers)
 
+daq_add_application(getAppsArguments get_apps_arguments.cxx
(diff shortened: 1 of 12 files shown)


plasorak:
And of course, the merge with `develop` left some traces. Sigh.

gcrone:
This does all seem rather convoluted. Since run control is the only thing that needs the command line parameters and it's written in Python I would have thought that it would be simpler to do the whole thing in Python without involving C++ dal methods.
  Adding dummy OpMon infrastructure_applications seems like an extra over complication. Wouldn't it be simpler to just extract the `Session environment` and use whatever is defined for `INFO_SVC`? You could even substitute `{APP_NAME}` for the `app.id`.
  Also, instead of having the control port hardwired in a `Service` for all the apps, how about just having a `BASE_PORT` variable and just increment it for each app? Even better if it could be passed to the drunc shell as an argument.

plasorak:
I agree with the first point, although I thought that's what we wanted to have, following discussions last Thursday.

For your second point, see the discussion here: https://dunescience.slack.com/team/UA6F5JMTQ

> Also, instead of having the control port hardwired in a Service for all the apps, how about just having a BASE_PORT variable and just increment it for each app? Even better if it could be passed to the drunc shell as an argument. 

Problem: in this case, the process manager would hand port numbers out, and the controller doesn't know how to talk to its applications because the port numbers are not accessible by it.

gcrone:
> For your second point, see the discussion here: https://dunescience.slack.com/team/UA6F5JMTQ
>
Not exactly sure what that link is supposed to take me to. When I tried clicking on it I just got your Slack profile

> > Also, instead of having the control port hardwired in a Service for all the apps, how about just having a BASE_PORT variable and just increment it for each app? Even better if it could be passed to the drunc shell as an argument.
> 
> Problem: in this case, the process manager would hand port numbers out, and the controller doesn't know how to talk to its applications because the port numbers are not accessible by it.

Isn't it the shell that parses the OKS database and would allocate ports rather than the process manager? Anyway, If the shell/process manager/controllers all parse the same database in the same way, why wouldn't they come up with the same set of ports?

plasorak:
Sorry, wrong Slack link: https://dunescience.slack.com/archives/C058BKQNUAZ/p1712244533762189

> Isn't it the shell that parses the OKS database and would allocate ports rather than the process manager? Anyway, If the shell/process manager/controllers all parse the same database in the same way, why wouldn't they come up with the same set of ports?

Indeed it's the shell (unified shell, or process manager shell), but the controller still does not have access to that piece of information.

Humm, that would involve parsing the whole configuration in each part of the run control, starting from the top segment and recursing into all the segments in the same way, which I think is likely to confuse users (and cause issues if there are disabled segments, for example). I agree it's not ideal...

gcrone (approved):
OK

PR #52, opened 2024-04-24
Add DUNEDAQ_DB_PATH to boot.json and fix data request queue rule for …

gcrone:
…CRP4 readout app

### DIFF
--- a/test/config/CRP4-session.data.xml
+++ b/test/config/CRP4-session.data.xml
@@ -88,7 +88,6 @@
  <attr name="application_name" type="string" val="daq_application"/>
  <rel name="runs_on" class="VirtualHost" id="vlocalhost"/>
  <rel name="modules">
-  <ref class="HermesController" id="hermes_10_10_5"/>
   <ref class="HermesController" id="hermes_10_10_4"/>
   <ref class="HermesController" id="hermes_10_10_3"/>
   <ref class="HermesController" id="hermes_10_10_2"/>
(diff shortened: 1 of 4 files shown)


jcfreeman2:
Running locally worked with a build of this feature branch; see the comment starting with "Using the code..." in https://github.com/DUNE-DAQ/daq-buildtools/pull/276 for details

jcfreeman2:
I'll leave any official review or comment on the data request rule fix for @alessandrothea ...

test/config/CRP4-session.data.xml
```diff
@@ -88,7 +88,6 @@
  <attr name="application_name" type="string" val="daq_application"/>
  <rel name="runs_on" class="VirtualHost" id="vlocalhost"/>
  <rel name="modules">
-  <ref class="HermesController" id="hermes_10_10_5"/>
```
alessandrothea: Why was the 6th wib endpoint removed?

gcrone: Because there are only 5 defined in dromap_CRP4.data.xml. Which is probably the file that needed fixing!!

gcrone: OK, I have put the 6th controller back in CRP4-session and added it to dromap_CRP4. I have also made a PR in oksconfgen to fix the dromap generation so that a regenerated xml will also have the 6th controller.

alessandrothea (approved):

PR #62, opened 2024-05-24
Make all the control port number 0

plasorak:
Requires https://github.com/DUNE-DAQ/restcmd/pull/24 so the apps can register themselves on the application connectivity server.

Still draft.


eflumerf:
I believe this PR is obsolete.

PR #66, opened 2024-05-31
Thread pinning FSM Action

PawelPlesniak:
With pre-conf, post-conf, and post-start. Requires:
Coredal https://github.com/DUNE-DAQ/coredal/pull/24


gcrone (approved):
Looks OK

PR #73, opened 2024-06-14
Roland sipos/oks wibmod

roland-sipos:


test/config/CRP4-session.data.xml
```diff
@@ -83,8 +83,49 @@
...
+ <attr name="reg_4" type="u8" val="59"/>
+ <attr name="reg_24" type="u8" val="223"/>
+ <attr name="reg_25" type="u8" val="51"/>
+ <attr name="reg_26" type="u8" val="137"/>
+ <attr name="reg_27" type="u8" val="103"/>
+ <attr name="reg_29" type="u8" val="39"/>
+ <attr name="reg_30" type="u8" val="39"/>
+</obj>
+
+<obj class="DaqApplication" id="hermes-CRP4">
```
gcrone, line 104: There is a WIECApplication called hermes-CRP4 in this file. I don't think we can have 2 applications with the same name. Although we can't yet test the WIECApplication as the module generating code doesn't exist yet, one should be removed or renamed.

schema/appmodel/fdmodules.schema.xml
```diff
@@ -192,6 +221,40 @@
   <attribute name="expected_timestamp_step" type="s32" init-value="-999" is-not-null="yes"/>
  </class>
 
+ <class name="WIBConf">
+  <attribute name="wib_addr" type="string" init-value="tcp://192.168.121.1:1234" is-not-null="yes"/>
```
gcrone, line 225: I don't know how the WIB addressing works but what if any is the relationship between this address and the ip_address in the `NICInterface` associated with the `HermesdataSender`? Should this also be a relationship to a `NICInterface`?

config/appmodel/moduleconfs.data.xml
```diff
@@ -160,14 +185,13 @@
...
  <attr name="emulation_mode" type="bool" val="1"/>
- <attr name="eal_args" type="string" val=""/>
  <rel name="emulation_conf" class="StreamEmulationParameters" id="stream-emu"/>
 </obj>
 
 <obj class="NICReceiverConf" id="def-nic-receiver-conf">
  <attr name="template_for" type="class" val="NICReceiver"/>
  <attr name="emulation_mode" type="bool" val="0"/>
- <attr name="eal_args" type="string" val=""/>
+ <attr name="eal_args" type="string" val="-l 0,1,3,5,7,59,61,63"/>
```
gcrone, line 193: The `eal_args` value looks suspiciously like the `used_lcores` in `DPDKPortConfiguration`. Should they be derived from there rather than entered here separately?

test/config/CRP4-session.data.xml
roland-sipos: Yes, this is a residual object that I wanted to keep for a similar behavior for testing. I didn't hear from the WIECApplication since today afternoon. I'll remove this offending app.

schema/appmodel/fdmodules.schema.xml
roland-sipos: No. Absolutely not. This config and module has nothing to do with the `NICInterface`. These URIs are front-end specific and are (SHOULD) not relate to any NICInterface addresses. This is the control path, not the data reception. I'd be reluctant to not have a separation between the control plane and data path addressing. I'd strongly advocate against mixing this. We can/should have a discussion about it at some point.

config/appmodel/moduleconfs.data.xml
roland-sipos: This is yet another "tuning" artifact in the readout. With kernel isolation and CPU masking EAL does not have visibility on CPU cores to use that are isolated by the kernel. So this is a per-host and pinning related config that now is required by the application itself. Yes, it is very bad and crappy... but I did add it, so the readout app at least can configure. We need to seriously rethink how we want to handle this. This is required by DPDK. We cannot ommit it. Maybe when we have a proper hardware representation in OKS we can generate it. Until then, unfortunately, this is what it is.

roland-sipos: Also, please mind the difference between `used_lcores` and this string argument list: 0 and 1 are added on top of the used lcores CPU set.

config/appmodel/moduleconfs.data.xml
```diff
@@ -273,4 +296,46 @@
  <attr name="prescale" type="u32" val="1000"/>
 </obj>
 
+<obj class="WIBConf" id="wib1001-conf">
```
glehmannmiotto, line 299: These OKS objects (WIBConf) should not be in the generic moduleconfs file, but, either in a separate file or in the CRP4 session file.

glehmannmiotto:
merged via another PR

PR #76, opened 2024-06-17
Cleaning up the CRP4 session config for use with drunc

alessandrothea:
- geoid renamed with meaningful ids
- session name fixed
- intorduced top root controller


glehmannmiotto (approved):
we should choose a naming conventions for geoid and the let dacqonf generate correctly. for now it's fine to use this suggestion, although very long.
One may go simply for detector name - crate#-slot#-stream-#

alessandrothea:
Agreed

alessandrothea:
I guess there are two options:
1. Ids with the geoparameters in them, but duplicating the content of the GeoID object
2. Or human-interpretable ids complementing the GeoID information

PR #80, opened 2024-06-18
WIECApplication module generation

alessandrothea:
Added the implementation of `WIECApplication::generate_module` based on detector-to-daq connection objects.
The generation algorithm searches for endpoints belonging to the same control host, and uses the list to generate both `WIBconfigurator` and `HermesModule` objects.

The wib/hermes modules will be generated only if the corresponding conf object associated to `WIECApplication` is not null.

The CRP4 configuration has been updated accordingly. Currently the configuration fails because of hermes timeouts when both hermes and wibs are enabled due to the WIB ipbus UDP server failing to reply within the timeout while the WIB logic configuration is being applied

### DIFF
--- a/config/appmodel/wiecconfs.data.xml
+++ b/config/appmodel/wiecconfs.data.xml
@@ -63,7 +63,7 @@
 
 <oks-data>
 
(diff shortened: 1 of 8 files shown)


glehmannmiotto (approved):
the generate modules test looks good.
Only suggestion is to give the WIECApplication a different name in the CRP4 example configuration. It is now hermes-CRP4; wiec-crp4 would be nicer.

alessandrothea:
Thanks, I meant to do change the app name, and now it's done.

PR #82, opened 2024-06-21
Gcrone/config fixes

gcrone:
Updates for the test-session configuration

### DIFF
--- a/config/appmodel/data-store-params.data.xml
+++ b/config/appmodel/data-store-params.data.xml
@@ -63,7 +63,7 @@
 
 <oks-data>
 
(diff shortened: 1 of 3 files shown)


eflumerf (approved):
LGTM, config runs in NanoRC on my VM. Do we want to add a session that doesn't use Kafka for testing with DRUNC when not at CERN?

gcrone:
It is not possible to use `drunc` without kafka, it uses it interbally, not just for `opmon` and `ers`. I would like to add a session or at least a set of objects that can be used in the existing test session without too much editing but I have failed to get a pocket-kafka container working which I think is the only way of using `drunc` outside EHN1.

PR #83, opened 2024-06-21
Thea/readout gen fix

gcrone:


alessandrothea (approved):
Looks FLAWLESS to me :P

PR #86, opened 2024-06-28
realised a bug when running fsm-test because it was trying to call el…

TiagoTAlves:
…isa even though it is only a prod action


PawelPlesniak (approved):
Tested with full FSM conf

PR #87, opened 2024-07-02
Mroda/ctb

mroda88:


eflumerf:
@mroda88 Is this stale?

mroda88:
It is. It was created last year when at some point it seemed it was supposed to be done. 
It could be closed but as I have to do it ASAP, I planned to savage whatever I did. 

If you prefer closing it, then go ahead and I'll open a new one once there is some work done.

eflumerf:
If there is still work planned, then we can keep it open for that work. What is the target release?

mroda88:
The target is 5.3.? (not the first one for the new TP format, but the next one). Will I make it? 
We'll find out.

PR #90, opened 2024-07-10
Disable TimeSyncs in def-hsi-handler and remove redundant trigger_can…

gcrone:
…didate_sink from RandomTCMakerModule schema. Also remove redundant commandline_parameters from some apps

Should also remove clock rate from RandomTCMakerConf and get it from DetectorConfig


eflumerf (approved):
Worked for me in integrationtest

PR #92, opened 2024-07-12
Re-introducing bespoke signal TRAP for gunicorn/connectiviy server

alessandrothea:
After the removal from drunc in https://github.com/DUNE-DAQ/drunc/pull/144

### DIFF
--- a/test/config/test-session.data.xml
+++ b/test/config/test-session.data.xml
@@ -115,9 +115,11 @@
 
 
 <obj class="ConnectionService" id="local-connection-server">
- <attr name="application_name" type="string" val="gunicorn"/>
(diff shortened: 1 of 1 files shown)


test/config/test-session.data.xml
```diff
@@ -115,9 +115,11 @@
 
 
 <obj class="ConnectionService" id="local-connection-server">
- <attr name="application_name" type="string" val="gunicorn"/>
+ <attr name="application_name" type="string" val="echo"/>
```
gcrone, line 118: Was this change just for debug, surely it should be `gunicorn` rather than `echo`?

gcrone (approved):
All seems good

PR #99, opened 2024-08-06
Updating trigger HSISourceModel with v4 changes

ArturSztuc:
Full description in https://github.com/DUNE-DAQ/trigger/pull/332. Do not merge without https://github.com/DUNE-DAQ/trigger/pull/332 approved.


gcrone (approved):
LGTM

glehmannmiotto (approved):
LGTM, although it is not very intuitive to have a simple "HSI2TC translator" apply a prescaling factor.

ArturSztuc:
Thank you @glehmannmiotto ! This prescaling was implemented for v4 after explicit requests, but I agree that it needs re-thinking anyway. E.g. now we prescale on whole HSIEvents. Should pre-scaling instead be done on the individual bits, or the output TC-types? I feel like different people might have different opinions on this, so I will add this to the list of tasks/discussions for the trigger (maybe next DS meeting).

PR #102, opened 2024-08-15
Updated print_detailed_config_info to show info for all apps

bieryAtFnal:
Recall that the `print_detailed_config_info` utility is an evolution of the `generate_modules_test` utility, and it initially added the print out of configuration parameters.

This latest round of changes includes the following:

- printout of full-system details without module generation
- printout of the modules for *all* DAQApplications with the configuration parameters included.

There are likely further enhancements to allow a user to focus on certain types of details, but I don't have a good sense yet of what those subsets might be, so it currently prints out everything.

Sample Usage:  `print_detailed_config_info test-session test/config/test-session.data.xml`

### DIFF
--- a/test/apps/print_detailed_config_info.cxx
+++ b/test/apps/print_detailed_config_info.cxx
@@ -135,7 +135,7 @@ main(int argc, char* argv[])
   // std::cout << "=====" << std::endl;
 
   std::cout << "++++++++++" << std::endl;
-  std::cout << "Full-system details without generation" << std::endl;
+  std::cout << "Full-system details without module generation" << std::endl;
   std::cout << "++++++++++" << std::endl;
   std::cout << std::endl;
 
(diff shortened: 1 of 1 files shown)


glehmannmiotto (approved):
I ran the print utility on the np02-oks-session (2 CRPs).
real	0m1.750s
user	0m1.115s
sys	0m1.185s
The output is polluted by some LOG outputs that should be demoted to DEBUG:
appmodel/src/WIECApplication.cpp:70
appmodel/src/ReadoutApplication.cpp:134
appmodel/src/ReadoutApplication.cpp:238
appmodel/src/ReadoutApplication.cpp:300
appmodel/src/ReadoutApplication.cpp:374
appmodel/src/ReadoutApplication.cpp:377
appmodel/src/DFApplication.cpp:148

The output is uploaded here: [output.txt](https://github.com/user-attachments/files/16633589/output.txt)
With 24k lines for a configuration with just 2 CRPs it may be not be obvious to use this dump for larger configurations. If the goal is to check specific configuration changes, it may be useful, int the future, to add command line options e.g. to allow specifying the segment or application to print out, or to optionally exclude queues, connections or not expand DetectorToDAQ connections.

bieryAtFnal:
That's good feedback on how to make this utility more useful in the future.

PR #104, opened 2024-08-22
Temporary Change To LB Capacity.

aeoranday:
This should be used as a workaround for testing development and not something that is kept into production. A true solution is still being investigated.

Following the minimal working example on https://github.com/DUNE-DAQ/daqconf/wiki/Setting-up-a-dunedaq-v5.1.0-Development-Area:
* Before this change, `ru-01` has a segmentation fault that is caused by a WIBEth frame expansion;
* After this change, the segmentation fault is gone and the expansion succeeds.

### DIFF
--- a/config/appmodel/moduleconfs.data.xml
+++ b/config/appmodel/moduleconfs.data.xml
@@ -145,7 +145,7 @@
 </obj>
 
 <obj class="LatencyBuffer" id="def-latency-buf">
- <attr name="size" type="u32" val="139008"/>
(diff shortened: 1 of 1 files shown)


bieryAtFnal (approved):
LGTM.  With this change, I can run the simple system defined in the appmodel sample config on the daq.fnal.gov teststand computer at Fermilab.

aeoranday:
#105 has a more well-rounded solution to the NUMA awareness problems people are facing, so I believe that PR trumps this one.

PR #105, opened 2024-08-22
choose numa unaware latency buffer

glehmannmiotto:
for a generic example, it is better to not rely on numa awareness and intrinsic allocator.


bieryAtFnal (approved):
LGTM. With this change, I can run the simple system defined in the appmodel sample config on the daq.fnal.gov teststand computer at Fermilab

PR #109, opened 2024-08-26
Converted TLOG() calls in several src files to TLOG_DEBUG(6) to avoid…

bieryAtFnal:
… spurious LOG statements when running tools like `print_detailed_config_info`.

While reviewing an earlier PR (number 102), Giovanna noticed that there were several `TLOG` statements in `appmodel` source code that cluttered up the screen output from some of the DB diagnostic tools.  

This PR changes those to `TLOG_DEBUG(6)` statements.

Please let me know if a different `TLOG` level than 6 would be better...


glehmannmiotto (approved):
LGTM

PR #111, opened 2024-08-28
Triggeralgs v4 to v5 port

ArturSztuc:
Porting trigger algorithms from v4 to v5.
Full description in  https://github.com/DUNE-DAQ/triggeralgs/pull/78.

Has to be merged together with https://github.com/DUNE-DAQ/triggeralgs/pull/78, https://github.com/DUNE-DAQ/trigger/pull/339, https://github.com/DUNE-DAQ/trgdataformats/pull/41 and https://github.com/DUNE-DAQ/fdreadoutlibs/pull/220.

### DIFF
--- a/schema/appmodel/trigger.schema.xml
+++ b/schema/appmodel/trigger.schema.xml
@@ -247,12 +247,12 @@
  <class name="TAMakerADCSimpleWindowAlgorithm">
   <superclass name="TAAlgorithm"/>
   <attribute name="adc_threshold" description="Threshold for the ADC integral algorithm" type="u32" init-value="10000" is-not-null="yes"/>
-  <attribute name="window_length" description="Window length (in ticks) for the ADC integral algorithm" type="u32" init-value="10000" is-not-null="yes"/>
+  <attribute name="window_length" description="Window length (in ticks) for the algorithm" type="u32" init-value="10000" is-not-null="yes"/>
  </class>
 
  <class name="TAMakerHorizontalMuonAlgorithm">
   <superclass name="TAAlgorithm"/>
-  <attribute name="window_length" description="Window length (in ticks) for the ADC integral algorithm" type="u32" init-value="10000" is-not-null="yes"/>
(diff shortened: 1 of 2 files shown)


schema/appmodel/trigger.schema.xml
```diff
@@ -247,11 +247,12 @@
  <class name="TAMakerADCSimpleWindowAlgorithm">
   <superclass name="TAAlgorithm"/>
   <attribute name="adc_threshold" description="Threshold for the ADC integral algorithm" type="u32" init-value="10000" is-not-null="yes"/>
-  <attribute name="window_length" description="Window length (in ticks) for the ADC integral algorithm" type="u32" init-value="10000" is-not-null="yes"/>
+  <attribute name="window_length" description="Window length (in ticks) for the algorithm" type="u32" init-value="10000" is-not-null="yes"/>
```
eflumerf, line 250: Is window_length a property of all TAAlgorithms? Should this attribute be moved to the base class?

ArturSztuc: Hi @eflumerf, it is not. Most of our algorithms do use a window_length, but not all of them (e.g. bundle algorithm, prescale algorithm, DBSCAN too I think), which is why I didn't add it to the base class.

aeoranday: @ArturSztuc Could there be another class level, e.g., `TAWindowAlgorithm` that is derived from `TAAlgorithm` and has `window_length`?

ArturSztuc: @aeoranday I actually don't know, but I imagine this should be possible

aeoranday (commented):
I notice many of the TCMakers are missing. Was this an oversight?

ArturSztuc:
@aeoranday That was definitely an oversight, I fixed it now in the last 2 commits.

aeoranday (approved):
LGTM.

Tested all related PRs by going through a test-session for a few different TAMaker and TCMaker configurations. All produced meaningful TAs/TCs, and the debug BundleN makers produced consistent counts for their configuration (reminds me to port `trgtools`).

PR #113, opened 2024-08-29
SmartDaqApplication uses SmartActionPlan

eflumerf:


glehmannmiotto (commented):
As commented in the corresponding PR for confmodel I would consider taking this opportunity for moving SmartDaqApplication, *ConnectionRule and *ConnectionDesciptor to confmodel.

glehmannmiotto:
So this PR does not modify any file wrt develop. It can be closed without merging, right?

PR #114, opened 2024-08-30
Add New TPG Configs

aeoranday:
These are the new configuration objects to be used the modular TPG design. Testing requires PRs in `fdreadoutlibs` and `tpglibs` with details in the associated `fdreadoutlibs` PR.

There is in issue in creating a new object when it is the first one in the class list https://github.com/DUNE-DAQ/dbe/issues/31. This may affect any testing for that creates new `AVXAbsRunSumProcessor` objects.

### Associated PRs
* https://github.com/DUNE-DAQ/fdreadoutlibs/pull/221
* https://github.com/DUNE-DAQ/tpglibs/pull/3


glehmannmiotto (commented):
there is no object of type AVXThresholdProcessor with ID tpg-threshold-proc.

glehmannmiotto:
after having added an opject in moduleconfs the system runs correctly. 

`
<obj class="AVXThresholdProcessor" id="tpg-threshold-proc">
 <attr name="plane0" type="u16" val="100"/>
 <attr name="plane1" type="u16" val="100"/>
 <attr name="plane2" type="u16" val="100"/>
</obj>
`
The test was done with a local merge of appmodel, fdreadoutlibs into the branches glm/post_proc (see branch glm/tpglibs for those 2 packages).
The next step will be to be able to enable/disable TP generation by plane and generate as TP handlers only for the planes for which TP generation is enabled (one source ID per TP handler will be needed).

glehmannmiotto:
closing PR since the branch is included into another PR

PR #116, opened 2024-08-30
simplify enabling/disabling of post processing tasks

glehmannmiotto:
this schema modification allows to raise the enabling and disabling of post-processing tasks at a higher level in the configuration. The change is meant to simplify enabling/disabling of processing  (hit finding) which is one of the things that is being turned on/off often these days.


schema/appmodel/application.schema.xml
```diff
@@ -339,8 +338,9 @@
   <superclass name="ResourceSetAND"/>
   <superclass name="SmartDaqApplication"/>
   <attribute name="tp_source_id" description="Source ID for TP handler if TPs are being generated" type="u32"/>
-  <attribute name="ta_source_id" description="Source ID if TAs are being generated" type="u32"/>
   <attribute name="application_name" type="string" init-value="daq_application" is-not-null="yes"/>
+  <attribute name="hit_finding_enabled" type="bool" init-value="true"/>
+  <attribute name="tp_processing_enabled" type="bool" init-value="true"/>
```
alessandrothea: What is the difference between tp_processing and hit finding?

alessandrothea: Only `hit_finding` seems to be used. If there's supposed to be only one, may I dare to suggest `tp_generation` as a name? It better relates to other `tp` entities while I don't think hitfinder is used anywhere else.

src/ReadoutApplication.cpp
```diff
@@ -328,7 +328,7 @@ ReadoutApplication::generate_modules(conffwk::Configuration* config, const std::
   conffwk::ConfigObject tp_queue_obj;
   conffwk::ConfigObject tpreq_queue_obj;
 
-  if (tph_conf) {
+  if (tph_conf && get_hit_finding_enabled()) {
```
alessandrothea: Probably a good precaution to add an error if the processing is enabled but no configuration is defined?

src/ReadoutApplication.cpp
```diff
@@ -405,7 +407,7 @@ ReadoutApplication::generate_modules(conffwk::Configuration* config, const std::
       dlh_outs.push_back(&ts_net_obj);
     }
 
-    if (tph_conf) {
+    if (tph_conf && get_hit_finding_enabled()) {
```
alessandrothea: Add sanity check here?

src/MLTApplication.cpp
```diff
@@ -346,7 +346,7 @@ MLTApplication::generate_modules(conffwk::Configuration* confdb,
         //   }
         // }
       }
-      if (ro_app->get_tp_source_id()!= 0) {
+      if (ro_app->get_tp_source_id()!= 0 && ro_app->get_hit_finding_enabled()) {
```
alessandrothea: Is this the right statement to detect that TPG is enabled in a readout application?
Shouldn't the `tph_conf` pointer be checked as well?
And if that's the case, shouldn't the `tp_source_id` be included in the other checks?

alessandrothea (commented):
The enabling logic should be checked to make sure inconsistent configurations (i.e. tp generation enabled + no processor configuration) are correctly identified as erroneous.
The role of the tp source id in the enabling logic should also be clarified.
Maybe worth considering the introduction of a ReadoutApplication helper method to combine these conditions.

src/ReadoutApplication.cpp
glehmannmiotto: I will make the existence of a conf object mandatory if tp_generation_enabled, and fire a FATAL error for the readout application if this is not the case.

schema/appmodel/application.schema.xml
glehmannmiotto: hit finding is the processing of raw data to generate TPs.
tp processing is the processing of TPs to generate TAs.
I can/will  change hit finding to TP generation, indeed.

src/ReadoutApplication.cpp
glehmannmiotto: I will add a single check at the start of the modules generation utility.

src/MLTApplication.cpp
glehmannmiotto: For MLT and other locations where the existence of hit finding is checked I will only use the tp_generation_enabled flag. The source_id == 0 check was really just a hack (and actually there is no requirement for a source id to not be 0).

glehmannmiotto:
I think I addressed all main points. This branch is already compatible with the eflumerf/ActionPlans_DaqModulesGroup branch of confmodel and appfwk

PR #122, opened 2024-09-10
Remove the variables that should be set by the run control

plasorak:
Requires: https://github.com/DUNE-DAQ/drunc/pull/237


plasorak:
Not needed anymore, now that configuration has been moved to daqsystemtests.

PR #127, opened 2024-09-16
Correct location for OpMon objects

mroda88:
The previous configuration had OpMon objects defined in the session that were referenced by the segments. This PR should make things right.


glehmannmiotto (approved):
LGTM ad should be merged when matching PRs are made on daqconf, daqsystemtest, listrev, ...

PR #132, opened 2024-09-24
Update application.schema.xml

glehmannmiotto:
removed obsolete attribute tp_source_id from ReadoutApplication


bieryAtFnal (approved):
I have run the latest tpstream_writing_test automated regression test with this change, and I've seen no problems, so this change looks good to me.

I also ran drunc interactively using the sample config in appmodel, and that did not see any adverse effects of this change.

PR #133, opened 2024-09-24
Go through DFApplications in DFOApplication and ReadoutApplication to

### LINKED ISSUE #130: Possible missing network connections
This may or may not be a problem, but it's possible that network connections are missing either in `test-session.data.xml` or in the way it gets processed. Say I loop over the applications in `test-session.data.xml` and run `generate_modules_test` on them, i.e.:
```
for app in hsi-to-tc-app hsi-01 tc-maker-1 mlt ru-01 ru-02 df-01 df-02 df-03 dfo-01 tp-stream-writer ; do generate_modules_test test-session $app $APPMODEL_SHARE/test/config/test-session.data.xml ; done
```
then I see things like, for `df-01`: 
```
inputs: ("trigger_decision_df-01@NetworkConnection", "fragments_df-01@NetworkConnection")
```
but I don't see a corresponding `trigger_decision_df-01@NetworkConnection` or `fragments_df-01@NetworkConnection` output anywhere.

eflumerf:
retrieve TriggerDecision inputs (DFO) and Fragment inputs (RU) to the DFApplication to completely specify all connections in the system during generation.

Resolves #130

### DIFF
--- a/src/DFOApplication.cpp
+++ b/src/DFOApplication.cpp
@@ -10,20 +10,21 @@
 
 #include "ModuleFactory.hpp"
 
-#include "conffwk/Configuration.hpp"
-#include "oks/kernel.hpp"
-#include "confmodel/Connection.hpp"
-#include "confmodel/NetworkConnection.hpp"
+#include "appmodel/DFApplication.hpp"
 #include "appmodel/DFOApplication.hpp"
 #include "appmodel/DFOConf.hpp"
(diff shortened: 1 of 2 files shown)


jcfreeman2:
When I create a `test-session` level plot using the feature branch, I now see formerly origin-less trigger decisions and fragments coming into DFApplications now coming from the DFO and readout applications. Approved.

PR #138, opened 2024-10-04
Adding python bindings for missing module generator helpers

alessandrothea:
That is 
- `FakeHSIApplication`
- `MLTApplication`
- `HSIEventToTCApplication`
- `WIECApplication`

Also added a convenience method to automatically select the correct generator given a python smart app object


gcrone (approved):
LGTM

Issue #135, opened 2024-10-01
Could the create_config_plot app be made to handle configs with more than one OKS Session in them?


bieryAtFnal:
I tried to run the `create_config_plot` app on the `example-configs.data.xml` file that is available in the `daqsystemtest` repo, and the app complained that it didn't like the fact that there was more than one Session object in the file.

```
(dbt) [biery@daq daqsystemtest]$ create_config_plot -f ./example-configs.data.xml -r local-1x1-config
2024-Oct-01 16:23:49,986 FATAL [appmodel::GraphBuilder::GraphBuilder(...) at /home/nfs/biery/dunedaq/01OctFDDevBleedingEdgeTest/sourcecode/appmodel/src/GraphBuilder.cpp:81] A graph tool error occured: Did not find one and only one Session instance in "./example-configs.data.xml" and its includes
```

Would it be possible to enhance the app so that it could handle configs with multiple OKS Sessions defined?  Of course, we would likely need to specify the OKS Session...

jcfreeman2:
Handled via PR #139

PR #139, opened 2024-10-08
JCF: Issue #135: add a session argument to create_config_plot to allo…

jcfreeman2:
…w it to disambiguate when there are multiple sessions in a database

To learn how to use it, read the updated documentation currently on the feature branch at
https://github.com/DUNE-DAQ/appmodel/tree/johnfreeman/issue135_multiple_sessions?tab=readme-ov-file#creating-a-diagram-of-your-configuration

### DIFF
--- a/apps/create_config_plot.cxx
+++ b/apps/create_config_plot.cxx
@@ -45,7 +45,7 @@ int main(int argc, char* argv[])
 
   ( "file,f", bpo::value<std::string> ( &oksfilename ), "OKS database file name" )
 
-  ( "root-object,r", bpo::value<std::string>(&object_uid), "OKS object UID of root vertex; must be session, segment or application")
+    ( "root-object,r", bpo::value<std::string>(&object_uid)->default_value(""), "OKS object UID of root vertex; must be session, segment or application")
     ( "session,s", bpo::value<std::string> ( &sessionname )->default_value(""),
     "Name of the session associated with the root object (only needed if >1 session in the database)" )
     ( "output,o", bpo::value<std::string> ( &outputfilename )->default_value("config.dot"),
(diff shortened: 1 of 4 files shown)


bieryAtFnal:
I ran some tests, and here are some notes:

- `create_config_plot -f config/daqsystemtest/example-configs.data.xml -s local-1x1-config -r local-1x1-config` produces the expected result (a plot of the 1x1 config)
- `create_config_plot -f config/daqsystemtest/example-configs.data.xml -s local-1x1-config -r local-2x3-config` reports no error and produces a plot of the 1x1 config.  This was a little unexpected, since I expected some sort of warning or error.
- `create_config_plot -f config/daqsystemtest/example-configs.data.xml -s local-2x3-config -r local-1x1-config` results in a crash (that is a little hard to notice among the other output from the app).  Here is the message from the console:
    - `create_config_plot: /home/nfs/biery/dunedaq/08OctFDDevBleedingEdgeTest3/sourcecode/appmodel/src/GraphBuilder.cpp:415: void appmodel::GraphBuilder::find_objects_and_connections(const ConfigObject&): Assertion `!m_objects_for_graph.contains(object.UID())' failed.`
- `create_config_plot -f config/daqsystemtest/example-configs.data.xml -s local-2x3-config -r local-2x3-config` produces the expected result (a plot of the 2x3 config)

Was I doing something wrong in the 2nd and 3rd tests?

jcfreeman2:
For 2 at least, the problem is that the algorithm searched _all_ objects in the database and needs to be narrowed to only search objects associated with the session. I'll make the needed fixes for (2) and (3) tomorrow morning; in particular (3) should be a cleaner, more informative error than an assertion fault.

jcfreeman2:
OK, I've updated the code. To get into details, now what happens is the code builds an internal, preliminary graph with the requested session as the root node, and checks that the actual requested root node is in the preliminary graph before proceeding as before.

bieryAtFnal:
Thanks.  I've confirmed that the commands that I listed earlier now either work or give a clear error message.

Unrelated to all of that, the ability to skip the "-s" argument when there is only one session in the database is nice.  Could there be support for a scenario where the "-r" argument is optional, namely when users *do* specify the "-s" argument and they would like the "seesion" root object to be plotted?

jcfreeman2:
Added support for defaulting the root object to the session name; updated documentation [here](https://github.com/DUNE-DAQ/appmodel/tree/johnfreeman/issue135_multiple_sessions?tab=readme-ov-file#creating-a-diagram-of-your-configuration)

bieryAtFnal (approved):
Looks great to me.
Thanks, John.

PR #143, opened 2024-10-14
Asztuc/mlt v4 to v5

ArturSztuc:
Porting MLT changes from v4 to v5, fully explained in https://github.com/DUNE-DAQ/trigger/pull/348


eflumerf (approved):
Builds, unit test passes, regression tests pass

PR #150, opened 2024-10-21
Reworking TC-type configs

ArturSztuc:
PR explained in: https://github.com/DUNE-DAQ/trigger/pull/351

### DIFF
--- a/schema/appmodel/trigger.schema.xml
+++ b/schema/appmodel/trigger.schema.xml
@@ -329,8 +329,6 @@
   <attribute name="ignore_tc" type="u32" is-multi-value="yes"/>
   <attribute name="use_bitwords" type="bool" init-value="false" is-not-null="yes"/>
   <attribute name="trigger_bitwords" type="string" is-multi-value="yes"/>
-  <attribute name="use_roi_readout" type="bool" init-value="false" is-not-null="yes"/>
-  <attribute name="use_readout_map" type="bool" init-value="false" is-not-null="yes"/>
   <relationship name="roi_group_conf" class-type="ROIGroupConf" low-cc="zero" high-cc="many" is-composite="no" is-exclusive="no" is-dependent="no"/>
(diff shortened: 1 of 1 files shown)


aeoranday (commented):
I noticed that `CustomTCMakerConf` still uses the integer type design instead of the new string one here. Is this an oversight?

https://github.com/DUNE-DAQ/appmodel/blob/develop/schema/appmodel/trigger.schema.xml#L130

ArturSztuc:
@aeoranday 
> I noticed that `CustomTCMakerConf` still uses the integer type design instead of the new string one here. Is this an oversight?
> 
> https://github.com/DUNE-DAQ/appmodel/blob/develop/schema/appmodel/trigger.schema.xml#L130

A little bit of an oversight: `CustomTCMaker` wasn't being used for anything in v5 tests, and it ended up being severaly out-of date, thanks for pointing this out.
There's more than just the TC-types to resolve here:
1. TC-types as you noticed
2. Replacing `trigger_intervals` with `trigger_rate_hz`
3. Removal of `clock_frequency_hz`
4. Complete rework of `CustomTCMaker` in a similar way to what was done to `RandomTCMaker` in https://github.com/DUNE-DAQ/trigger/pull/350

I'm happy to do this, but probably in a separate PR. For now `CustomTCMaker` is so out of date that probably shouldn't be used...

In fact, perhaps we should completely wipe out `CustomTCMaker` from existance. `RandomTCMaker` has configurable TC-type and TC-window output, and we can have multiple `RandomTCMaker`s running concurrently... making `CustomTCMaker` a bit redundant.

aeoranday:
The argument for removing `CustomTCMaker` is sensible, and I agree that it would be out-of-scope for this PR. Though, I imagine it will remain a low priority and end up taking a while before removing.

aeoranday (approved):
LGTM.

ArturSztuc:
Thank you Alex. I was going to go through trigger issues page and resolve as many as possible over the next ~2 weeks, and I see that `CustomTCMaker` features in a few of them - I'll work on that next, so it won't be forgotten :)

PR #154, opened 2024-11-07
Descriptions in trigger schema & small cleanup

ArturSztuc:
This PR goes together with:
1. https://github.com/DUNE-DAQ/trigger/pull/361
2. https://github.com/DUNE-DAQ/daqsystemtest/pull/154

Changes:
1. Added descriptions to the trigger schemas
3. Removed few that are obsolete/not used attributes/objects
     - This includes HSI passthrough. It was already removed in v4, and somehow it was not removed in v5. This included removal of the hsi-passthrough code in the trigger repository (see the trigger PR above)
3. Changed the default value for `TCDataProcessor::td_readout_limit` from `100` to `1,000,000`: it used to be in terms of milliseconds, now it is in terms of ticks.

Tested by running integration tests (all pass), and the example test configuration.

### DIFF
--- a/schema/appmodel/trigger.schema.xml
+++ b/schema/appmodel/trigger.schema.xml
@@ -127,9 +127,9 @@
  <class name="CustomTCMakerConf">
   <superclass name="StandaloneTCMakerConf"/>
   <attribute name="template_for" type="class" init-value="CustomTCMaker"/>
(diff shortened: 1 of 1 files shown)


glehmannmiotto (approved):
LGTM

PR #159, opened 2024-11-26
Disable postprocessing option in TriggerAplication

ArturSztuc:
We can now disable the post-processing in `TriggerApplication`, so we can e.g. disable TC-making (or TA-making, if we're running TA making on the same machines as the tc-making).

EDIT: this PR is standalone now. <del>This PR has to go together with  https://github.com/DUNE-DAQ/daqsystemtest/pull/167</del>

Tested by running with both `tc_generation_enabled` set to 1 and 0, and seeing if the TCs with `kPrescale` type are present in the output hdf5 file.

### DIFF
--- a/schema/appmodel/trigger.schema.xml
+++ b/schema/appmodel/trigger.schema.xml
@@ -385,6 +385,8 @@
   <superclass name="ResourceSetAND"/>
   <superclass name="SmartDaqApplication"/>
   <attribute name="application_name" type="string" init-value="daq_application" is-not-null="yes"/>
+  <attribute name="ta_generation_enabled" description="Enable TA generation as a post-processing task" type="bool" init-value="true"/>
+  <attribute name="tc_generation_enabled" description="Enable TC generation as a post-processing task" type="bool" init-value="true"/>
   <relationship name="data_subscriber" class-type="DataReaderConf" low-cc="one" high-cc="one" is-composite="no" is-exclusive="no" is-dependent="no"/>
   <relationship name="trigger_inputs_handler" class-type="DataHandlerConf" low-cc="one" high-cc="one" is-composite="no" is-exclusive="no" is-dependent="no"/>
   <method name="generate_modules" description="Generate daq module dal objects for streams of thie TriggerApplication on the fly">
--- a/schema/appmodel/trigger.schema.xml
+++ b/schema/appmodel/trigger.schema.xml
@@ -385,8 +385,8 @@
   <superclass name="ResourceSetAND"/>
   <superclass name="SmartDaqApplication"/>
   <attribute name="application_name" type="string" init-value="daq_application" is-not-null="yes"/>
-  <attribute name="ta_generation_enabled" description="Enable TA generation as a post-processing task" type="bool" init-value="true"/>
-  <attribute name="tc_generation_enabled" description="Enable TC generation as a post-processing task" type="bool" init-value="true"/>
+  <attribute name="ta_generation_enabled" description="Disable TA generation as a post-processing task -- only to be done by the experts for e.g. TP performance studies" type="bool" init-value="true"/>
+  <attribute name="tc_generation_enabled" description="Disable TC generation as a post-processing task -- only to be done by the experts for e.g. TA performance studies" type="bool" init-value="true"/>
   <relationship name="data_subscriber" class-type="DataReaderConf" low-cc="one" high-cc="one" is-composite="no" is-exclusive="no" is-dependent="no"/>
   <relationship name="trigger_inputs_handler" class-type="DataHandlerConf" low-cc="one" high-cc="one" is-composite="no" is-exclusive="no" is-dependent="no"/>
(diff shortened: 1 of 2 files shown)


bieryAtFnal:
I believe that it would be helpful to include some text in the description of these new configuration parameters to indicate that they are not intended for production use.  This would help non-expert users understand more about their intention.

Something along the lines of "Enable TA generation as a post-processing task. This is for performance testing only, and it should only disabled by knowledgeable experts."

ArturSztuc:
Hi @bieryAtFnal, thank you for the suggestion. I just pushed a commit with an expanded description, please let me know what you think.

bieryAtFnal:
Looks good, thank you.

Another minor comment is that naming the config params tx_generation_disabled might help convey the sense that "enabled" is the normal condition.  It may be too late to change that, though.

ArturSztuc:
I think having it as `tx_generation_disabled` would make a lot of sense, but it then would not be consistent with how the same thing is already done in the readout [HERE](https://github.com/DUNE-DAQ/appmodel/blob/4c38cf7ad1d58dcbf130356d74849d271e46c41b/schema/appmodel/application.schema.xml#L361-L362).

So I would be happy to change this to `tx_generation_disabled` if we also change this in the readout application (I'm happy to do it myself, but probably need @glehmannmiotto to weigh in on this).

glehmannmiotto:
I think that the readout and trigger applications cases are a bit different.
In a readout application there is processing of raw data for TP generation and, within the same application, it may make sense to generate TAs (processing one plane of 1 detector unit for each TA maker).
I don't think that it is equally sensible for a trigger application that generates TAs (i.e. it receives TPs) to generate TCs (which by definition should combine TAs from different detector regions). Therefore, I think that @bieryAtFnal 's suggestion to have a single attribute and call it tx_generation_enabled makes sense (and is a good idea) for a TriggerApplication, while I wouldn't change the ReadoutApplication.

ArturSztuc:
Thank you @glehmannmiotto (and @MRiganSUSX for reminding me about this PR, I dropped a ball...)...
@bieryAtFnal  I just pushed a commit ([cb53d27](https://github.com/DUNE-DAQ/appmodel/pull/159/commits/cb53d273b2415230ca3bc1c5dfe78e924f014874)) that makes the requested change.

MRiganSUSX:
Thanks @ArturSztuc. I mostly appreciate the cleanup in TriggerApplication source, making use of the `create_network_connection` (that was there but unused). 

I can confirm that:
- there is now new variable for the trigger application, that controls tx generation, false by default
- with it at its default value we get TAs and TCs, just as before
- with it set to true, TAs and TCs are no longer being produced

Some questions: 
- this variable only affects the trigger flow TX objects (TPs coming from readout and their associated handlers), and not TCs from standalone makers (Random) or HSIEvents (Timing). So it is 'only' turning off the readout TP flow TX generation. I assume this is what is wanted for the tests, but mentioning it here if we ever wanted to expand to control to all TX objects. 
- the current implementation treats this variable as a 'global' switch, and turns off both TA and TC making together. Would we ever want to have the higher granularity option of controlling TA and TC making separately?

ArturSztuc:
Thanks @MRiganSUSX ! 
> Some questions:
> 
> * this variable only affects the trigger flow TX objects (TPs coming from readout and their associated handlers), and not TCs from standalone makers (Random) or HSIEvents (Timing). So it is 'only' turning off the readout TP flow TX generation. I assume this is what is wanted for the tests, but mentioning it here if we ever wanted to expand to control to all TX objects.

Yes, so there would be a few things to be done here: 
1. Disabling the `StandaloneMaker`'s TC-making capabilities. I would say we probably don't need that, since we can simply comment them out/take them off the `standalone_candidate_maker_confs` in `MLTConf`. This would have been more tricky with the "self-triggered" chain, hece the new variable.
2. This change is mostly targeted towards the `DataProcessors` objects, and in line with what was done in the readout: the idea is that the data processor is still receiving objects, buffering them, just not making the next object afterwards. With this in mind, I definitely agree that we could include it into the `TCDataProcessor` too, so it receives all the TCs, but doesn't make any TDs. I thought about implementing it, but decided against it for now: MLT already has a lot of logic that can achieve this, and we probably want to completely re-write MLT anyway (at which point we should accommodate this feature).

So to conclude, I think we should include this in `TCDataProcessor`/`MLT` when we start re-writing/cleaning up the MLT

> * the current implementation treats this variable as a 'global' switch, and turns off both TA and TC making together. Would we ever want to have the higher granularity option of controlling TA and TC making separately?

`TriggerApplication` can only generate either TAs or TCs, but not both at the same time -- so to have TA-making in the `TriggerApplication` (currently done in `ReadoutApplication`), we would have two TriggerApplications in TriggerSegment -- one making TAs, one making TCs (or even more wild, we would have one TriggerApplication per APA for TA-making). So this new variable, `tx_generation_disable` is actually acting the way you want, you could switch it to `true` for either TA-making `TriggerApplication`, or TC-making `TriggerApplication`, or both. That was also what Giovanna's comment above was alluding to.

PR #161, opened 2024-12-09
Add TP Time Over Threshold Minima Configurables.

aeoranday:
Incoming signals will have a width (time over threshold) that is greater than incoming noise. I've added a configurable filter that passes TPs with a width that is at least the configured value.

This filter becomes more useful at lower signal amplitude thresholds, but it will require signal shape knowledge to make an accurate estimate on what may be a reasonable amplitude-width threshold combination.

I've made the range of values `[1..65535]`, but the upper limit in this range is essentially infinite. I expect typical values to be in the range `[1..20]`.

## Linked PRs
* https://github.com/DUNE-DAQ/tpglibs/pull/11,
* https://github.com/DUNE-DAQ/fdreadoutlibs/pull/241,
* https://github.com/DUNE-DAQ/daqsystemtest/pull/168.

## Testing
Completing the usual system tests in `daqsystemtest` and a local session test run with various `tot_minimum_plane<n>` values.
* At a TOT minimum of 1, this should be the same as it has been.
* At a TOT minimum > 1, the default emulated signals are suppressed since they are a width of 1.
* At a TOT minimum > 1 and using `/cvmfs/dunedaq.opensciencegrid.org/assets/files/b/6/8/wibeth_single_channel.bin`, the TP `time_over_threshold / 32` ranges from the configured minimum to 63. These values are expected for the pedestal estimation and shape of the emulated signal (rectangle of amplitude = 666 and width = 63).

Completed a low amplitude threshold online test with NP04. Resultant TPs are as expected and the system was able to keep up for reasonable amplitude-width threshold combinations.


glehmannmiotto (approved):
LGTM. I would have naively expected the ToT parameters to be part of the specific ProcessingStep, actually.

aeoranday:
If it were its own `ProcessingStep`, then it could make more sense. Maybe this motivates a new category of processing steps that check the state and send out TPs depending on different criteria. However, there may be a problem when trying to combine criteria in the same way as the "body" processing steps due to input/output format and complexity.

PR #162, opened 2024-12-16
switched to using the TPWriter instance index (currently always zero)…

bieryAtFnal:
… in the TPStreamWriter data filename in TPWriterApplication, instead of the source_id, as discussed at the Core Software meeting last week.

The goal of this is to have the raw-data and tp-stream-data filenames be more consistent.

### DIFF
--- a/src/TPWriterApplication.cpp
+++ b/src/TPWriterApplication.cpp
@@ -84,10 +84,11 @@ TPStreamWriterApplication::generate_modules(conffwk::Configuration* confdb,
     throw(BadConf(ERS_HERE, "No SourceIDConf given to TPWriterApplication!"));  
   }
 
+  uint tpw_idx = 0;
   std::string tpwrUid("tpwriter-"+std::to_string(source_id->get_sid()));
   confdb->create(dbfile, "TPStreamWriterModule", tpwrUid, tpwrObj);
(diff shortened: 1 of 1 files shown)


src/TPWriterApplication.cpp
```diff
@@ -84,10 +84,11 @@ TPStreamWriterApplication::generate_modules(conffwk::Configuration* confdb,
     throw(BadConf(ERS_HERE, "No SourceIDConf given to TPWriterApplication!"));  
   }
 
+  uint tpw_idx = 0;
   std::string tpwrUid("tpwriter-"+std::to_string(source_id->get_sid()));
   confdb->create(dbfile, "TPStreamWriterModule", tpwrUid, tpwrObj);
   tpwrObj.set_by_val<uint32_t>("source_id", source_id->get_sid());
-  tpwrObj.set_by_val("writer_identifier", fmt::format("{}_tpw_{}", UID(), source_id->get_sid()));
+  tpwrObj.set_by_val("writer_identifier", fmt::format("{}_tpw_{}", UID(), tpw_idx));
```
eflumerf, line 91: I'm not sure I trust this. Why is the SourceIDConf for the TPWriters always 0? Shouldn't it just be set to different values when there are more than one TPW in the configuration?

eflumerf (approved):
Sorry, I was thought the problem was that the SourceID was always zero; got the problem backwards.

PR #163, opened 2024-12-18
Added the fragment network output connection to the MLT and Trigger apps...

bieryAtFnal:
… for completeness and to match what is included in the Readout app.

I discovered that these connections were missing from the MLT and Trigger apps when trying to add "sender is ready" code to the `datahandlinglibs` `DefaultRequestHandlerModel` code, and I noticed that there were no fragment output connections in the MLT and Trigger app configs.  I need these output connections to be defined in order for my changes to check if Fragment senders are ready in `do_start()` methods to be fully beneficial.

The absence of these connections is nicely shown in the config diagrams that we can generate with the create_config_plot utility.  

The changes that I've made for this PR successfully provide this missing connection config information.  However, I can image a few follow-up actions:

- pull out the copy-pasted common code that looks up the fragment network connection into a function
- run the clang-format utility on the xyzApplication.cpp code
- check with Artur if there are ramifications from the disable_tx_generation config flag on the handling of DataRequests and sending of Fragments by Trigger apps.

I propose to take care of those in separate Issues and Pull Requests.

I can provide instructions for testing these changes, both by generating config plots and running regression tests, if desired.

Attached are sample diagrams that illustrate the problem and the fix.  The first diagram shows the Readout app config from the `minimal_system_quick_test` (regression test) configuration.  It **does** have the `fragments_df-01` external connection.  The second diagram shows the MLT app from the MSQT test before the fix.  It does **not** show the `fragments_df-01` connection.  The third diagram shows the MLT config after the fix, and it does show the `fragments_df-01` connection.

17Dec_msqt_ruapp
17Dec_msqt_mlt
18Dec_msqt_mlt

### DIFF
--- a/src/MLTApplication.cpp
+++ b/src/MLTApplication.cpp
@@ -51,6 +51,7 @@
 #include "appmodel/ReadoutApplication.hpp"
 #include "appmodel/TriggerApplication.hpp"
 #include "appmodel/appmodelIssues.hpp"
+#include "appmodel/DFApplication.hpp"
 
 #include "appmodel/StandaloneTCMakerConf.hpp"
 #include "appmodel/StandaloneTCMakerModule.hpp"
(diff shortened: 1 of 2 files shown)


eflumerf (approved):
Used `generate_modules_test` to show the before/after difference, double-checked `minimal_system_quick_test` as well.

PR #165, opened 2025-01-13
 Added the fragment network output connections to the FakeHSI app…

bieryAtFnal:
… so that we can look them up in the FakeHSI init() method and ensure that those connections are ready to be used in the FakeHSI do_start() method.

Also, cleaned up some commented-out lines in the MLT and Trigger application code.

Here are the config diagrams for the FakeHSI app in the 2x3 example system config before and after the change.

13Jan_local-2x3_hsi-fake-01

13Jan_local-2x3_hsi-fake-01_v2


eflumerf (approved):
LGTM. On another channel, Alessandro has discussed the idea of moving some of these "other app" lookups into the Session (whatever) object, to try and preserve some degree of encapsulation.

PR #170, opened 2025-02-15
Added back a line in trigger.schema.xml that seems to have been lost …

bieryAtFnal:
…in recent merges.

I believe that this change will eliminate the errors that we've seen recently with the TP-related overnight regression tests.

I wasn't able to figure out which merge lost this line, but that doesn't seem terribly important...


bieryAtFnal:
I've re-confirmed that this addition improves the state of the overnight v5 build, so I'm going to go ahead and merge it.

PR #171, opened 2025-02-26
Gcrone/private session

gcrone:
This PR provides a configuration helper class with methods to extract connection descriptors and source ids from other applications in the session while keeping the actual Session object private. The helper object is created by the `appfwk::ConfigurationManager` and passed to a smart applications `generate_modules()` method.

Requires corresponding PR in `appfwk` (https://github.com/DUNE-DAQ/appfwk/pull/319)


eflumerf:
Setting to v5.5.0, at least for a determination of whether this is still relevant or the best way forward.

mroda88 (changes requested):
This breaks the CTB logic as we need to request connections for the CTB app. We need to find the right solution.

mroda88:
This will also fix #272

mroda88 (changes requested):
Before we forget, the MLT needs to be updated to pick up both source ids of the CTB. I'm happy to take care of this next week. Sorry for the oversight.

mroda88 (approved):
This is now ok, and it has been tested at EHN1.

PR #172, opened 2025-02-26
Use common object factory code to generate configuration objects

gcrone:
Add a helper class to create configuration objects for smart application `generate_modules()` methods. In this PR the code is accessed via the configuration helper introduced in #171 but the code is separate and an `ObjectFactory` could be instantiated without the helper.


alessandrothea (approved):
LGTM

gcrone:
Superseded by #194

PR #173, opened 2025-03-06
TP Replay Application (v5) 

MRiganSUSX:
Details here: https://github.com/DUNE-DAQ/trigger/pull/382


CMakeLists.txt
```diff
@@ -20,7 +20,8 @@ daq_oks_codegen(application.schema.xml fdmodules.schema.xml trigger.schema.xml w
   NAMESPACE dunedaq::appmodel DEP_PKGS confmodel)
 
 daq_add_library(ReadoutApplication.cpp SmartDaqApplication.cpp
-	DFApplication.cpp DFOApplication.cpp TPWriterApplication.cpp FakeDataApplication.cpp FakeHSIApplication.cpp DTSHSIApplication.cpp TriggerApplication.cpp MLTApplication.cpp HSIEventToTCApplication.cpp WIECApplication.cpp DaphneApplication.cpp
+	DFApplication.cpp DFOApplication.cpp TPWriterApplication.cpp FakeDataApplication.cpp FakeHSIApplication.cpp DTSHSIApplication.cpp TriggerApplication.cpp MLTApplication.cpp TPReplayApplication.cpp HSIEventToTCApplication.cpp WIECApplication.cpp 
+	DFApplication.cpp DFOApplication.cpp TPWriterApplication.cpp FakeDataApplication.cpp FakeHSIApplication.cpp DTSHSIApplication.cpp TriggerApplication.cpp TPReplayApplication.cpp MLTApplication.cpp HSIEventToTCApplication.cpp WIECApplication.cpp DaphneApplication.cpp
```
ArturSztuc: Why is this duplicated, with different order and second one containing `DaphneApplication.cpp` and the first one not? One copy of each should suffice:
```suggestion
	DFApplication.cpp DFOApplication.cpp TPWriterApplication.cpp FakeDataApplication.cpp FakeHSIApplication.cpp DTSHSIApplication.cpp TriggerApplication.cpp TPReplayApplication.cpp MLTApplication.cpp HSIEventToTCApplication.cpp WIECApplication.cpp DaphneApplication.cpp SocketSenderApplication.cpp
```

(I added `SockedSenderApplication.cpp` as it is needed in the development)

schema/appmodel/trigger.schema.xml
```diff
@@ -406,4 +406,40 @@
...
  </class>
 
+ <class name="TPReplayModule">
+  <superclass name="DaqModule"/>
+  <relationship name="configuration" class-type="TPReplayModuleConf" low-cc="one" high-cc="one" is-composite="no" is-exclusive="no" is-dependent="no"/>
+ </class>
+
+ <class name="TPReplayModuleConf">
+  <attribute name="template_for" type="class" init-value="TPReplayModule"/>
+  <attribute name="number_of_loops" type="u32" init-value="1" is-not-null="yes"/>
```
ArturSztuc: I haven't looked into the code yet, but is it also possible to "just keep on running" until user stops the drunc run?

schema/appmodel/trigger.schema.xml
```diff
@@ -406,4 +406,40 @@
...
+  <attribute name="maximum_wait_time_us" type="u32" init-value="1000" is-not-null="yes"/>
+  <attribute name="channel_map" type="string" init-value="PD2HDChannelMap" is-not-null="yes"/>
+  <attribute name="total_planes" type="u32" init-value="0" is-not-null="yes"/>
+  <relationship name="filter_out_plane" class-type="PlaneNumberConf" low-cc="zero" high-cc="many" is-composite="no" is-exclusive="no" is-dependent="no"/>
```
ArturSztuc: Does this need to be a relationship? Attributes can be vectors. It would make more sense if `PlaneNumberConf` was some more complex object, but it's just an integer...

Example of multi-valued attribute: https://github.com/DUNE-DAQ/appmodel/blob/bfe6b6cd2fdab1559eb2c3574c7cf3931395c7af/schema/appmodel/trigger.schema.xml#L307

There's even potentially a way of constraining this to a specific range with `range` keyword and either `0,1,2` or `0...2`.

schema/appmodel/trigger.schema.xml
```diff
@@ -406,4 +406,40 @@
...
+  <relationship name="tp_streams" class-type="TPStreamConf" low-cc="one" high-cc="many" is-composite="no" is-exclusive="no" is-dependent="no"/>
+</class>
+
+ <class name="PlaneNumberConf">
+  <attribute name="plane" type="u32" init-value="4294967295" is-not-null="yes"/>
+ </class>
```
ArturSztuc: Could be just removed and replaced with an attribute that can have multiple values.

ArturSztuc (commented):
Review in progress, I'll comment more when it runs (see PR in trigger).

schema/appmodel/trigger.schema.xml
```diff
@@ -406,4 +406,36 @@
...
+  <superclass name="ResourceSetAND"/>
+  <superclass name="SmartDaqApplication"/>
+  <attribute name="application_name" type="string" init-value="daq_application" is-not-null="yes"/>
+  <relationship name="tp_source_ids" class-type="SourceIDConf" low-cc="zero" high-cc="many" is-composite="no" is-exclusive="no" is-dependent="no"/>
+  <relationship name="tprm_conf" class-type="TPReplayModuleConf" low-cc="one" high-cc="one" is-composite="no" is-exclusive="no" is-dependent="no"/>
+  <relationship name="tp_handler" class-type="DataHandlerConf" low-cc="one" high-cc="one" is-composite="no" is-exclusive="no" is-dependent="no"/>
+  <method name="generate_modules" description="Generate daq module dal objects for TPReplayApplication on the fly">
+   <method-implementation language="c++" prototype="std::vector&lt;const dunedaq::confmodel::DaqModule*&gt; generate_modules(conffwk::Configuration*, const std::string&amp;, const confmodel::Session*) const override" body=""/>
+  </method>
+ </class>
```
ArturSztuc, line 439: Would be good to have a `description` for each of these. We're not super consistent with it in the rest of the `trigger.schema.xml`, but probably over 95% of items do have it.

ArturSztuc (approved):
LGTM!

The only comment I really have is about adding `description` to all the new items in `trigger.schema.xml`. We should have them, but feel free to add them after this PR.

I tested this too, although was not able to do as thorough testing as I wanted, simply because I cannot find enough TP data with v5. Ideally I would like multiple consecutive files (in time), from multiple APAs etc. and test various combinations. Instead, I tested 1 file per APA with few different algorithms.

So it works, more testing is needed, but we should merge this (after the other two PRs are approved) :)
Thanks for a lot of good work here!

ArturSztuc (approved):
LGTM

PR #176, opened 2025-03-19
Update in DAPHNE configuration

mroda88:


mroda88:
Tested with the Daphne today and it looks ok

PR #183, opened 2025-04-05
TdeAMCSender class

alessandrothea:
A TDE AMC sender descriptor needs to be added to the schema.
Added `TdeAmcDataSender` class to describe TDE AMC boards .

Part of DUNE-DAQ/daq-deliverables#166


roland-sipos (approved):
LGTM

PR #184, opened 2025-04-10
Boost ASIO socket reader

denizergonul:
A Boost.Asio-based socket reader plugin shall be implemented as a lightweight alternative to the current DPDK-based solution, which is overly complex for low-bandwidth electronic devices.


roland-sipos (approved):
LGTM

PR #193, opened 2025-05-02
Mroda/ctb v5.3

mroda88:
This PR adds configuration objects and establish links with other applications to include the CTB.


alessandrothea (approved):
LGTM

PR #194, opened 2025-05-06
ConfigObjectFactory

alessandrothea:
The module generation process implemented by SmartApplication’s generate_module in appmodel involves generating a number of application framework objects: modules, queues, network connections.
The solution is introducing a single `ConfigObjectFactory` utility class, that aggregates all generic helper functions in a single, coherent API. A ConfigObjectFactory is created at the start of generate_modules and used within the scope of module generation

Detailed description of the change: https://docs.google.com/document/d/1jD0W6Bb4iAD3O6Q8naCi4lbr9E8-lj5e/edit#heading=h.wu6wg8bdyyeh


alessandrothea:
After some intense debugging, the integration tests are now all green

src/DFApplication.cpp
```diff
@@ -183,6 +161,8 @@ DFApplication::generate_modules(conffwk::Configuration* confdb,
 {
   std::vector<const confmodel::DaqModule*> modules;
 
+  const auto obj_fac = ConfigObjectFactory(confdb, dbfile, UID());
```
gcrone: Should we use the simplified constructor in all these 'examples'?

src/FakeHSIApplication.cpp
```diff
@@ -153,47 +148,24 @@ FakeHSIApplication::generate_modules(conffwk::Configuration* confdb,
   // Time Sync network connection
   if (dlhConf->get_generate_timesync()) {
     std::string tsStreamUid = tsNetDesc->get_uid_base() + std::to_string(id);
```
gcrone: Should just be `std::to_string(id)`. The uid base from the descriptor is added by `create_net_obj`

gcrone: Should be fixed by the new commit https://github.com/DUNE-DAQ/appmodel/pull/194/commits/92b92693cd36dbc9f4c3e45d601f7482b69621d2

src/DFApplication.cpp
alessandrothea: We shall!

src/FakeHSIApplication.cpp
alessandrothea: Well spotted

jcfreeman2:
One thing that needs to be changed before we merge into `patch/fddaq-v.5.3.x` is `cmake/appmodelConfig.cmake.in`. Right now it breaks a test build, https://github.com/DUNE-DAQ/daq-release/actions/runs/15003678738, because it assumes use of a newer `daq-cmake` version, `v3.1.0`, than the one we've been using for the patch builds (`v3.0.5`). The newer daq-cmake version includes changes to the build interface.

alessandrothea:
> One thing that needs to be changed before we merge into `patch/fddaq-v.5.3.x` is `cmake/appmodelConfig.cmake.in`. Right now it breaks a test build, https://github.com/DUNE-DAQ/daq-release/actions/runs/15003678738, because it assumes use of a newer `daq-cmake` version, `v3.1.0`, than the one we've been using for the patch builds (`v3.0.5`). The newer daq-cmake version includes changes to the build interface.

Fixed.

gcrone (approved):
OK, All good now

PR #196, opened 2025-05-14
Adding the factory to remaining applications

alessandrothea:
No actual change to the code, but added to remaining application for consistency (the factory acts as a OksFile (un)lock guard.


gcrone (approved):
LGTM

PR #204, opened 2025-05-29
Specialization of ReadoutApplication to use for TDE readout

ShyamB97:
Due to the TDE crate reading data from both CRPs, a readout application that is capable of reading out multiple CRPs/APAs is required to avoid larger issues downstream of the readout.


mroda88 (approved):
LGTM

PR #213, opened 2025-06-12
Sort TP streams in NP02 readout app

ShyamB97:
The NP02 readout application will readout multiple detector elements, so (for now), sort the TP streams by the major number of the source ID of the TP input and detector stream (TPs produced from detector stream 5XX will end up in TP handlers 5X).


ShyamB97:
Tested last week by taking a run with BDE, inspecting the tp stream files and verifying that TPs from CRP4 were written from TPHandler 4x, and TPs from CRP5 were written from TPHandler 5x.

alessandrothea (approved):
LGTM

PR #216, opened 2025-06-23
PDS/TPC RawDataProcessor added to the appmodel trigger schema

ramarine:
RawDataProcessor changed into TPCRawDataProcessosr and PDSRawDataProcessor. 
This is a simple way to have configurable thresholds and masked channels for the PDSRawDataProcessor in a similar way as th WIBs. 

Code passed all integration tests when run only with patch/fddaq-v5.3.x for the other repos.

### DIFF
--- a/schema/appmodel/trigger.schema.xml
+++ b/schema/appmodel/trigger.schema.xml
@@ -177,7 +177,6 @@
 
  <class name="PDSRawDataProcessor" description="A class which contains info about PDS. For example here you can add threshold for specific channels.">
   <superclass name="RawDataProcessor"/>
-  <attribute name="rawdataprocessor_id" type="u32" init-value="234" is-not-null="yes"/>
   <attribute name="default_adc_intg_thresh" type="u32" init-value="0" is-not-null="yes"/>
  </class>
 
--- a/schema/appmodel/trigger.schema.xml
+++ b/schema/appmodel/trigger.schema.xml
@@ -175,6 +175,14 @@
   <relationship name="configuration" class-type="MLTConf" low-cc="one" high-cc="one" is-composite="no" is-exclusive="no" is-dependent="no"/>
  </class>
 
+ <class name="PDSRawDataProcessor" description="A class which contains info about PDS. For example here you can add threshold for specific channels.">
+  <superclass name="RawDataProcessor"/>
+  <attribute name="rawdataprocessor_id" type="u32" init-value="234" is-not-null="yes"/>
+  <attribute name="default_adc_intg_thresh" type="u32" init-value="0" is-not-null="yes"/>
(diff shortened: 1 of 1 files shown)


schema/appmodel/trigger.schema.xml
```diff
@@ -175,6 +175,14 @@
   <relationship name="configuration" class-type="MLTConf" low-cc="one" high-cc="one" is-composite="no" is-exclusive="no" is-dependent="no"/>
  </class>
 
+ <class name="PDSRawDataProcessor" description="A class which contains info about PDS. For example here you can add threshold for specific channels.">
+  <superclass name="RawDataProcessor"/>
+  <attribute name="rawdataprocessor_id" type="u32" init-value="234" is-not-null="yes"/>
+  <attribute name="default_adc_intg_thresh" type="u32" init-value="0" is-not-null="yes"/>
+  <relationship name="processing_steps" class-type="ProcessingStep" low-cc="zero" high-cc="many" is-composite="no" is-exclusive="no" is-dependent="no"/>
```
alessandrothea: `processing_steps` and `sot_minima` are `TPCRawDataProcessor` specific field.
Remove from `PDSRawDataProcessor` and `RawDataProcessor`

schema/appmodel/trigger.schema.xml
alessandrothea, line 1: It seems that some confusin crept up in the schema compared to what was in `thea/pds_processor_config`.
In particular remove `processing_steps`  and `sot_minima` are `TPCRawDataProcessor` specific attributes and should exist only in there.

alessandrothea (changes requested):
See comments.

schema/appmodel/trigger.schema.xml
ramarine: They need to be in the RawDataProcessor because if not, the code crashes on the WIB part. Maybe that need to be investigated on its own.

schema/appmodel/trigger.schema.xml
ramarine: All clear!

alessandrothea:
@ramarine please change the title to some more explicative

schema/appmodel/trigger.schema.xml
```diff
@@ -175,6 +175,12 @@
   <relationship name="configuration" class-type="MLTConf" low-cc="one" high-cc="one" is-composite="no" is-exclusive="no" is-dependent="no"/>
  </class>
 
+ <class name="PDSRawDataProcessor" description="A class which contains info about PDS. For example here you can add threshold for specific channels.">
+  <superclass name="RawDataProcessor"/>
+  <attribute name="rawdataprocessor_id" type="u32" init-value="234" is-not-null="yes"/>
```
wesketchum: `rawdataprocessor_id` should be removed.

wesketchum (changes requested):

wesketchum (approved):

alessandrothea (approved):

PR #218, opened 2025-06-24
Add capacity parameter to NetworkConnectionDescription.

eflumerf:
Set capacity of Fragment output connection to twice the number of streams (i.e. enough for just 2 TriggerRecords) in ReadoutApplication.

This capacity is then passed to `ipm`, see https://github.com/DUNE-DAQ/ipm/pull/109


mroda88 (approved):
Tested at EHN1

PR #220, opened 2025-07-07
Added the timesync input connection to the Fake HSI object in the FakeHSIApplication

bieryAtFnal:
…, so that it will correctly appear in our configuration diagrams and other diagnostic tools.

We've noticed that the TimeSync "connection" into the FakeHSI application is missing in our configuration diagrams.  The change in this PR fixes that.  I'll add before/after diagrams to this PR.

To see the results of these changes, we can run 

- `generate_modules_test local-1x1-config hsi-fake-01 config/daqsystemtest/example-configs.data.xml`
    - and look at the "input objects"

or run

- `create_config_plot -f config/daqsystemtest/example-configs.data.xml -s local-1x1-config`, etc

This PR will have a follow-up one in `hsilibs`, in which we make use of the TimeSync connection that is providing in the confguration...

07Jul_local-1x1-config_system
07Jul_local-1x1-config_system


eflumerf (approved):
LGTM. small_footprint_quick_test uses FakeHSI, verified that it still works.

PR #222, opened 2025-07-08
Gcrone/custom resource2

gcrone:
Use the new `Resource` schema from confmodel.
Requires https://github.com/DUNE-DAQ/confmodel/pull/73 to be merged


mroda88 (approved):
Tested with updated NP02 configurations and it worked, including runconf-ui

PR #226, opened 2025-07-23
Johnfreeman/issue225 remove resource code

### LINKED ISSUE #225: `TPReplayApplication` should be updated to work with new OKS resource scheme
As part of merging `fddaq-v5.3.3` into `develop`, I [needed to modify](https://github.com/DUNE-DAQ/appmodel/commit/bb054c8518f6a489a37c8dc7df074f1ed80e9e09) `TPReplayApplication` so it could build given the new resource scheme Gordon implemented on the v5.3.x patch branch. However, although it builds, it's not in a usable state; in particular `TPReplayApplication::contained_resources() ` will throw an exception informing users that it hasn't been implemented. A proper implementation of `TPReplayApplication::contained_resources()` by an expert should be performed.

jcfreeman2:


schema/appmodel/trigger.schema.xml
```diff
@@ -440,17 +440,13 @@
 
  <class name="TPReplayApplication">
   <superclass name="SmartDaqApplication"/>
-  <superclass name="ResourceSetDisableAND"/>
```
gcrone, line 443: `TPReplayApplication` does not use resources so shouldn't be a `ResourceSetDisableAND` but on second thoughts, should it be a `Resource` so that it can be disabled

MRiganSUSX: Thanks for the comments / suggestions @gcrone 
What is the advantage of this being a resource (apologies for my ignorance) ? It's an application so it can just be included / excluded as needed ?

gcrone: A `Resource` can be disabled by adding it to the disabled relationship of the `Session`. An `Application` cannot be disabled you would have to delete it from the list of applications in its `Segment` or add the `Segment` to the disabled list. If you never need to disable this application on its own then it doesn't need to be a `Resource`.

MRiganSUSX: That (to me) sounds like a definition problem? How do we distinguish resources from applications? 

Regardless, here are my findings:
- testing as is, there are no visible errors but there is no data flowing through the replay system. No TP processors are present it seems?
- adding `<superclass name="Resource"/>` makes it work as expected.
So it seems like it _has to_ be a resource?

MRiganSUSX: Just to sum up (so we can have a functional replay again), there is definitely some change needed as it does not work as expected as is. It did work for me with `<superclass name="Resource"/>` but I am unsure whether that is the best solution. Thoughts @gcrone ?

gcrone: This is a bug. It was supposed to work regardless whether or not `TPReplayApplication` is a `Resource`. I have opened an issue in confmodel https://github.com/DUNE-DAQ/confmodel/issues/79

MRiganSUSX: This now works as expected without the need to define it as a `Resource` with fix in https://github.com/DUNE-DAQ/confmodel/pull/80.

MRiganSUSX (approved):
This now works as expected without the need to define it as a Resource with fix in https://github.com/DUNE-DAQ/confmodel/pull/80.

PR #228, opened 2025-07-29
Removed the setting of the d2dconn in the reader module in CRTReaderApplication.cpp...

bieryAtFnal:
... since that module doesn't contain a socket reader or writer.


denizergonul (approved):
Thanks.

PR #229, opened 2025-07-30
Postprocessing delay min and max wait conf params

denizergonul:
- If data arrives less than `post_processing_delay_min_wait` milliseconds since the last time postprocessing eligibility check was done, postprocessing will be delayed. [This value was previously hardcoded as 1.](https://github.com/DUNE-DAQ/datahandlinglibs/blob/84bd8fc8ec8330c3cb1056dcbbcc63cc196051e5/include/datahandlinglibs/models/detail/DataHandlingModel.hxx#L279)

- If no data arrives in `post_processing_delay_max_wait` milliseconds, postprocessing eligibility check will still take place. This is a new feature of delayed postprocessing to cover the case when there is data in the buffer that is eligible for postprocessing but no new data arrives to trigger the check.

I would appreciate your input on the `init-value`s.

Related: [Delayed postprocessing rework #66](https://github.com/DUNE-DAQ/datahandlinglibs/pull/66)

### DIFF
--- a/schema/appmodel/application.schema.xml
+++ b/schema/appmodel/application.schema.xml
@@ -80,7 +80,7 @@
 
 <oks-schema>
 
-<info name="" type="" num-of-items="56" oks-format="schema" oks-version="862f2957270" created-by="gjc" created-on="thinkpad" creation-time="20230616T091343" last-modified-by="gjc" last-modified-on="latitude" last-modification-time="20250626T122830"/>
+<info name="" type="" num-of-items="56" oks-format="schema" oks-version="862f2957270" created-by="gjc" created-on="thinkpad" creation-time="20230616T091343" last-modified-by="dergonul" last-modified-on="np04-srv-028.cern.ch" last-modification-time="20250729T142105"/>
 
 <include>
  <file path="schema/confmodel/dunedaq.schema.xml"/>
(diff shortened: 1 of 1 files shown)


denizergonul:
Please feel free to add more people as reviewers.

ShyamB97:
Hi @denizergonul, For the `init-value`s, were you able to try running the system with different values to check the impact on the system? If so, then the values which resulted in the more stable performance could be used for the default values.

denizergonul:
During the tests we used only these values. 
```
post_processing_delay_ticks = 312500  # was already there
post_processing_delay_min_wait = 1    # was already hardcoded before
post_processing_delay_max_wait = 5
```
Do you think we should give it a try with different values as well? The current results do seem stable to me, though.

ShyamB97:
I think if the current values are stable then that should be sufficient, and we can use those for the initial values.

schema/appmodel/application.schema.xml
```diff
@@ -153,6 +153,8 @@
   <attribute name="input_data_type" description="Type of data received by this readout module" type="string" init-value="WIBEthFrame"/>
   <attribute name="generate_timesync" type="bool" init-value="true" is-not-null="yes"/>
   <attribute name="post_processing_delay_ticks" description="Number of clock tick by which post processing of incoming data shall be delayed." type="u64" init-value="0" is-not-null="yes"/>
```
ShyamB97, line 155: Don't you need to change this to 312500? As per the values you tested?

roland-sipos (approved):
LGTM

ShyamB97 (approved):
Changes tested with 04/08/25 nightly. Integration tests passed.

schema/appmodel/application.schema.xml
denizergonul: Delayed postprocessing is an optional feature; it is also valid when set to 0 and postprocessing is performed immediately after each insertion to the buffer.

PR #231, opened 2025-08-08
Make FDFakeReaderModule work with CRT frames

denizergonul:
- `CRTReaderConf` to have a common class for CRT frames:
```
<obj class="CRTReaderConf" id="def-emu-crt-receiver-conf">
 <attr name="template_for" type="class" val="FDFakeReaderModule"/>
 <attr name="emulation_mode" type="bool" val="1"/>
 <rel name="emulation_conf" class="StreamEmulationParameters" id="stream-emu"/>
</obj>
```
- Added `connections` since it was required when using `FDFakeReaderModule`

The following is the list of relevant branches (where `SkipListLatencyBufferModel` is used as the buffer type):
asiolibs: dte/asio_integtest
daqsystemtest: dte/emu_crt
appmodel: dte/fakereader_crt
fdreadoutlibs: dte/queue_for_crt
fdreadoutmodules: dte/queue_for_crt
crtmodules: dte/crt_integtest

### DIFF
--- a/schema/appmodel/fdmodules.schema.xml
+++ b/schema/appmodel/fdmodules.schema.xml
@@ -80,7 +80,7 @@
 
 <oks-schema>
 
-<info name="" type="" num-of-items="23" oks-format="schema" oks-version="862f2957270" created-by="gjc" created-on="thinkpad" creation-time="20230616T091343" last-modified-by="gjc" last-modified-on="latitude" last-modification-time="20250623T154337"/>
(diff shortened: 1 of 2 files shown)


wesketchum:
Hi @denizergonul , this looks good, but can you remind me of what I should run to do a test of this? Any other packages (with special branches) needed to be pulled down when I do that?

denizergonul:
The following is the list of branches where `BinarySearchQueueModel` is used as the buffer type:
asiolibs: dte/asio_integtest
daqsystemtest: dte/emu_crt
appmodel: dte/fakereader_crt
fdreadoutlibs: dte/queue_for_crt
fdreadoutmodules: dte/queue_for_crt

bieryAtFnal (approved):
I'm surprised to see that the setting of the "connections" on line 148 needed to be added back.  I want to look into why, but for now, I'll approve this PR because the `asiolibs` and `crtmodules` integtests are running successfully with these changes.

PR #234, opened 2025-08-11
Processor metric collection related configuration

xinyue-uoft:
### Summary of changes

This PR adds four new attribute definitions under classes `AVXFrugalPedestalSubtractProcessor` and `TPCRawDataProcessor`. 

These attributes are added to control behaviour of classes introduced in the following PR.

https://github.com/DUNE-DAQ/tpglibs/pull/27
https://github.com/DUNE-DAQ/fdreadoutlibs/pull/272
https://github.com/DUNE-DAQ/datahandlinglibs/pull/70

#### In `AVXFrugalPedestalSubtractProcessor`
- `metric_collect_data_sample_rate` defaults to 128, it is the rate at which metric data is stored to the dedicated metric buffer, waiting to be polled. The rate is defined in terms of time samples (or how many times `process` method is ran).
- `metric_collect_toggle_state` is a bool that controls whether the metric saving function is ran periodically at all (rate defined above). The default value is false (off).

#### In `TPCRawDataProcessor`
- `metric_collect_enable` is a bool toggle that controls whether upstream `ProcessorMetricCollector` object is created and called to collect metrics. The default is false (off). 
- `metric_collect_opmon_rate` controls the rate in terms of data frames (in terms of frame counter), for which the `ProcessorMetricCollector` updates its internal record of metric from polling from all processors registered on it. 

### Usage Note

It should be noted that it only makes sense to have `metric_collect_toggle_state` and `metric_collect_enable` in the two classes set to the same value (both true or both false), as it corresponds to the producer and collector, respectively.

### DIFF
--- a/schema/appmodel/trigger.schema.xml
+++ b/schema/appmodel/trigger.schema.xml
@@ -198,7 +198,6 @@
   <superclass name="RawDataProcessor"/>
   <relationship name="processing_steps" class-type="ProcessingStep" low-cc="one" high-cc="many" is-composite="no" is-exclusive="no" is-dependent="no"/>
   <relationship name="sot_minima" description="TP samples over threshold minimum requirement by plane." class-type="SamplesOverThresholdMinima" low-cc="one" high-cc="one" is-composite="no" is-exclusive="no" is-dependent="no"/>
-  <attribute name="metric_collect_enable" description="Whether to enable processor metric collection" type="bool" init-value="0"/>
   <attribute name="metric_collect_opmon_rate" description="The rate at which processor metric is polled from processors for monitoring" type="u32" init-value="1"/>
  </class>
 
--- a/schema/appmodel/trigger.schema.xml
+++ b/schema/appmodel/trigger.schema.xml
@@ -100,7 +100,7 @@
  <class name="AVXFrugalPedestalSubtractProcessor" description="TPG frugal pedestal subtraction signal processor. Frugally updates the pedestal estimation and outputs the difference with the input signal.">
   <superclass name="ProcessingStep"/>
   <attribute name="accum_limit" description="Accumulation limit to increment/decrement the estimated pedestal value." type="u8" init-value="10" is-not-null="yes"/>
-  <attribute name="metric_collect_data_sample_rate" description="Rate in terms of time samples to store metric values in processors." type="u8" init-value="128" is-not-null="no"/>
+  <attribute name="metric_collect_data_sample_rate" description="Rate in terms of time samples to store metric values in processors." type="u64" init-value="128" is-not-null="no"/>
   <attribute name="metric_collect_toggle_state" description="Whether metric storaging function is enabled." type="bool" init-value="0" is-not-null="no"/>
  </class>
 
--- a/schema/appmodel/trigger.schema.xml
+++ b/schema/appmodel/trigger.schema.xml
@@ -100,6 +100,8 @@
  <class name="AVXFrugalPedestalSubtractProcessor" description="TPG frugal pedestal subtraction signal processor. Frugally updates the pedestal estimation and outputs the difference with the input signal.">
   <superclass name="ProcessingStep"/>
   <attribute name="accum_limit" description="Accumulation limit to increment/decrement the estimated pedestal value." type="u8" init-value="10" is-not-null="yes"/>
+  <attribute name="metric_collect_data_sample_rate" description="Rate in terms of time samples to store metric values in processors." type="u8" init-value="128" is-not-null="no"/>
+  <attribute name="metric_collect_toggle_state" description="Whether metric storaging function is enabled." type="bool" init-value="0" is-not-null="no"/>
  </class>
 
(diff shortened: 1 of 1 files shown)


schema/appmodel/trigger.schema.xml
```diff
@@ -100,6 +100,8 @@
  <class name="AVXFrugalPedestalSubtractProcessor" description="TPG frugal pedestal subtraction signal processor. Frugally updates the pedestal estimation and outputs the difference with the input signal.">
   <superclass name="ProcessingStep"/>
   <attribute name="accum_limit" description="Accumulation limit to increment/decrement the estimated pedestal value." type="u8" init-value="10" is-not-null="yes"/>
+  <attribute name="metric_collect_data_sample_rate" description="Rate in terms of time samples to store metric values in processors." type="u64" init-value="128" is-not-null="no"/>
+  <attribute name="metric_collect_toggle_state" description="Whether metric storaging function is enabled." type="bool" init-value="0" is-not-null="no"/>
```
aeoranday: Wouldn't it be better to have these available to all `ProcessingStep` classes? Then a class may use the values if it actually needs them.

schema/appmodel/trigger.schema.xml
```diff
@@ -100,6 +100,8 @@
  <class name="AVXFrugalPedestalSubtractProcessor" description="TPG frugal pedestal subtraction signal processor. Frugally updates the pedestal estimation and outputs the difference with the input signal.">
   <superclass name="ProcessingStep"/>
   <attribute name="accum_limit" description="Accumulation limit to increment/decrement the estimated pedestal value." type="u8" init-value="10" is-not-null="yes"/>
+  <attribute name="metric_collect_data_sample_rate" description="Rate in terms of time samples to store metric values in processors." type="u64" init-value="128" is-not-null="no"/>
```
aeoranday: This is not a rate but a period. Time samples of what?

schema/appmodel/trigger.schema.xml
aeoranday: Would we want separate processors to feature different sampling periods? Does this configurable go higher to TPG level/`TPCRawDataProcessor`?

schema/appmodel/trigger.schema.xml
```diff
@@ -196,6 +198,8 @@
   <superclass name="RawDataProcessor"/>
   <relationship name="processing_steps" class-type="ProcessingStep" low-cc="one" high-cc="many" is-composite="no" is-exclusive="no" is-dependent="no"/>
   <relationship name="sot_minima" description="TP samples over threshold minimum requirement by plane." class-type="SamplesOverThresholdMinima" low-cc="one" high-cc="one" is-composite="no" is-exclusive="no" is-dependent="no"/>
+  <attribute name="metric_collect_enable" description="Whether to enable processor metric collection" type="bool" init-value="0"/>
```
aeoranday: In `fdreadoutlibs`, it appears that this configurable and `metric_collect_toggle_state` are not cross-checked. So is it possible to have this turned on with the other off and vice versa? In this case, are we adding unnecessary processing?

aeoranday (dismissed):

schema/appmodel/trigger.schema.xml
xinyue-uoft: > Would we want separate processors to feature different sampling periods? Does this configurable go higher to TPG level/`TPCRawDataProcessor`?

When a processor gets configured, it only sees these attributes belonging to its class. Then if we would like this configuration to go to a higher level, we would need to call some `configure_collection_rate` method, which is extra coupling.

aeoranday: This is the opposite of coupling. With the PR as it is, configuring TPG and the metric collection take the same route when they do two different things.

If it's on `ProcessingStep` with the option of `is-not-null="no"`, then this should avoid introducing another method, makes it available to all processors, and shouldn't break the current configs that do not set this value. Processors already only pull the values they need.

If it's on `TPCRawDataProcessor`, only the top level `TPGenerator` should see this and treat it separately from the base functionality of TPG.

xinyue-uoft:
### Updated to the metric collection enabling logic

There is now a single enable toggle under class `ProcessingStep` named `metric_collect_toggle_state`. Default is disabled and this attribute can be null.

At `TPGenerator` configuration, it will check through all processing steps. If and only if one of the processing steps has an enabled `metric_collect_toggle_state`, it will initialize and run the `ProcessorMetricCollector` object and thread. 

The benefit is that 1. This prevents the situation where metric collect at processors is enabled, but disabled at `ProcessorMetricCollector` level. 2. Allows setting individual enable state and collection rate, for each concrete type of `ProcessingStep`. 3. Overall make configuration easier.

schema/appmodel/trigger.schema.xml
xinyue-uoft: Fixed, only processor level toggle remains.

ShyamB97: should be resolved now

PR #236, opened 2025-09-02
Add TRMonReqApplication

eflumerf:
This application consists of a TRMonRequestorModule and a DataWriterModule, and is designed to request TriggerRecords from the DFApplications and write them to separate files (generally, should be in one-event-per-file mode) for fast-turnaround DQM.

Part of https://github.com/DUNE-DAQ/dfmodules/pull/439


alessandrothea (approved):
LGTM
Tested at EHN1

PR #239, opened 2025-09-10
Extension of the PDS schema for daphne mezzanine

mroda88:
The PR does the following

- Changes the schema to remove the json in the Daphne Configuration object in favour of the a structure of objects
- Extends the functionality of DaphneApplication to work with the both Daphne v2 and Daphne v3 mezzanine.


schema/appmodel/PDS.schema.xml
```diff
@@ -93,7 +93,9 @@
   <superclass name="SmartDaqApplication"/>
   <superclass name="ResourceSetDisableAND"/>
   <relationship name="configuration" class-type="DaphneConf" low-cc="one" high-cc="one" is-composite="no" is-exclusive="no" is-dependent="no"/>
-  <relationship name="detector_connections" description="List of detector components controlled by this application" class-type="FelixDetectorToDaqConnection" low-cc="zero" high-cc="many" is-composite="no" is-exclusive="no" is-dependent="no"/>
+  <relationship name="detector_connections" description="List of detector components controlled by this application" class-type="DetectorToDaqConnection" low-cc="zero" high-cc="many" is-composite="no" is-exclusive="no" is-dependent="no"/>
```
alessandrothea, line 96: the `low_cc` for both `detector_connections` and `hermes_modules_conf` is `zero`.
Does a `DaphneApplication` with those fields empty make sense?

schema/appmodel/PDS.schema.xml
```diff
@@ -170,16 +173,21 @@
...
   <superclass name="DaqModule"/>
-  <attribute name="address" description="ip address of the board" type="string" is-not-null="yes"/>
   <relationship name="daphne_conf" class-type="DaphneConf" low-cc="one" high-cc="one" is-composite="no" is-exclusive="no" is-dependent="no"/>
   <relationship name="board_conf" class-type="DaphneV2BoardConf" low-cc="one" high-cc="one" is-composite="yes" is-exclusive="yes" is-dependent="yes"/>
-  <method name="get_slot" description="">
-   <method-implementation language="C++" prototype="auto get_slot() const { return get_board_conf()-&gt;get_slot_id(); }" body="BEGIN_HEADER_PROLOGUE&#xA;#include &lt;appmodel/DaphneV2BoardConf.hpp&gt;&#xA;END_HEADER_PROLOGUE"/>
-  </method>
+ </class>
+
+ <class name="DaphneV2ControllerModule">
```
alessandrothea, line 185: What is the purpose of the `DaphneV2ControllerModule` and `DaphneV3ControllerModule` since they are identical to the parent class?

src/DaphneApplication.cpp
```diff
@@ -74,160 +94,133 @@ DaphneApplication::generate_modules(const confmodel::Session* session) const
...
 
-    // Loop over senders
-    for (const auto* felix_sender : det_senders) {
-
-      if ( felix_sender->is_disabled(*session) ) {
-        TLOG() << "Skipping disabled sender: " << felix_sender->UID();
-        continue;
-      }
+    if ( ! net_conn) {
+      if ( ! flx_conn ) throw BadConf(ERS_HERE, d2d_conn->UID() + " is neither felix or eth connection");
```
alessandrothea: why the nested if instead of a clearer `if ( !net_conn & !flx_conn) { ... }`?

alessandrothea:
LGTM in general.
A few small details should be looked into for consistency

schema/appmodel/PDS.schema.xml
mroda88: Yes, when it's connected to a v2 version. I'll open an issue to remind us that when daphne v2 is discontinued this code should be clean up. Among other things, this minimum multiplicity should change.

schema/appmodel/PDS.schema.xml
mroda88: They do different operations. From the point of view of the configuration they are the same, i.e. they need the same information. But the controllers do different things, see https://github.com/DUNE-DAQ/daphnemodules/pull/55

schema/appmodel/PDS.schema.xml
mroda88: here is the Issue https://github.com/DUNE-DAQ/appmodel/issues/257

src/DaphneApplication.cpp
mroda88: This has been changed according to the comment.

PR #241, opened 2025-09-23
Rma/wib fr count v2

ramarine:
Quick change for the issue marked here: https://github.com/DUNE-DAQ/fdreadoutlibs/issues/260

I added two variables in the TPCRawDataProcessor which can be configured. 
 
frame_count_limit -> defaulted at 100 as it was before. 
 tp_count_limit -> defaulted at at 210 TPs, 2x nominal TP count per dlh after 100 framse 
 
 Code passed self-made consistency checks. As in testing if the values are well collected from the configuration, if the TPs are send to the sink when frame/tp condition met.

### DIFF
--- a/schema/appmodel/trigger.schema.xml
+++ b/schema/appmodel/trigger.schema.xml
@@ -196,8 +196,8 @@
 
  <class name="TPCRawDataProcessor">
   <superclass name="RawDataProcessor"/>
-  <attribute name="frame_count_thr" description="The maximum number of frames before TPs are sent to the TP sink.. &#xA;" type="u32" init-value="100" is-not-null="yes"/>
-  <attribute name="TP_count_thr" description="The maximum number of TPs collected after some frames, before all the TPs are sent to the TP sink. " type="u32" init-value="400000000" is-not-null="yes"/>
+  <attribute name="frame_count_limit" description="The maximum number of frames before TPs are sent to the TP sink. &#xA;" type="u32" init-value="100" is-not-null="yes"/>
+  <attribute name="tp_count_limit" description="The maximum number of TPs, before the TPs are sent to the TP sink. " type="u32" init-value="400000000" is-not-null="yes"/>
   <relationship name="processing_steps" class-type="ProcessingStep" low-cc="one" high-cc="many" is-composite="no" is-exclusive="no" is-dependent="no"/>
   <relationship name="sot_minima" description="TP samples over threshold minimum requirement by plane." class-type="SamplesOverThresholdMinima" low-cc="one" high-cc="one" is-composite="no" is-exclusive="no" is-dependent="no"/>
   <attribute name="metric_collect_opmon_rate" description="The rate at which processor metric is polled from processors for monitoring" type="u32" init-value="1"/>
--- a/schema/appmodel/trigger.schema.xml
+++ b/schema/appmodel/trigger.schema.xml
@@ -196,6 +196,8 @@
 
  <class name="TPCRawDataProcessor">
   <superclass name="RawDataProcessor"/>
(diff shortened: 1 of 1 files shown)


schema/appmodel/trigger.schema.xml
```diff
@@ -196,6 +196,8 @@
 
  <class name="TPCRawDataProcessor">
   <superclass name="RawDataProcessor"/>
+  <attribute name="frame_count_limit" description="The maximum number of frames before TPs are sent to the TP sink. &#xA;" type="u32" init-value="100" is-not-null="yes"/>
```
aeoranday:
```suggestion
  <attribute name="frame_count_limit" description="The maximum number of frames before TPs are sent to the TP sink." type="u32" init-value="100" is-not-null="yes"/>
```

aeoranday (changes requested):
LGTM. There is an HTML newline that should be removed though.

schema/appmodel/trigger.schema.xml
```diff
@@ -175,8 +175,8 @@
 
  <class name="ProcessingStep" description="Base class for TPG processors.">
   <superclass name="Jsonable"/>
-  <attribute name="metric_collect_time_sample_period" description="If metric_collect_toggle_state is true, the processor will store values of its monitored quantities every metric_collect_time_sample_period time samples." type="u64" init-value="256" is-not-null="no"/>
-  <attribute name="metric_collect_toggle_state" description="Whether metric storaging function is enabled." type="bool" init-value="0" is-not-null="no"/>
+  <attribute name="metric_collect_time_sample_period" description="If metric_collect_toggle_state is true, the processor will store values of its monitored quantities every metric_collect_time_sample_period time samples." type="u64" init-value="256"/>
+  <attribute name="metric_collect_toggle_state" description="Whether metric storaging function is enabled." type="bool" init-value="0"/>
```
aeoranday: Why was this modified?

schema/appmodel/trigger.schema.xml
```diff
@@ -397,30 +386,32 @@
   <attribute name="time_after" description="Time, in ticks, after the HSI timestamp, for the TriggerCandidate end_time" type="u32" init-value="1001" is-not-null="yes"/>
  </class>
 
+ <class name="TPCRawDataProcessor">
+  <superclass name="RawDataProcessor"/>
+  <attribute name="frame_count_limit" description="When this number of frames is reached the TPs are sent to the sink. " type="u32" init-value="100" is-not-null="yes"/>
```
aeoranday: As I understood from the feature request from @ShyamB97, `frame_count_limit` should also be toggle-able, similar to how you have done with `tp_count_limit`.

aeoranday (changes requested):
I noticed a lot of out-of-scope changes in one commit. Many of them seem like a movement for alphabetical order, so I suppose it's fine.

schema/appmodel/trigger.schema.xml
ramarine: for the moment we only want the TP toggleable such that we can get back the previous behaviour. 
Could be implemented in a future PR.

schema/appmodel/trigger.schema.xml
ramarine: i re-added is-not-null="no" to the two variables...

schema/appmodel/trigger.schema.xml
roland-sipos: It's the same corresponding issue though. Please check my comments on the draft PR in the corresponding repo.

roland-sipos (changes requested):
Address the toggling of both limits.

alessandrothea (changes requested):
Please implement changes request by Alex and Roland

aeoranday (approved):
Thanks for making the changes. LGTM.

alessandrothea (approved):
LGTM

roland-sipos (approved):

Issue #247, opened 2025-10-07
Review DaqApplication and SmartDaqApplication relationship


gcrone:
Consider using inheritance to implement the relationship between `DaqApplication` and `SmartDaqApplication` (and potentially the name of the latter).
   * Use `generate_modules` to populate the `modules` list in `DAQApplication` (using `Configuration::update(...)` to apply the changes). 
   * Simplify `generate_modules` signature returning `void`

**Affected packages**: `appmodel`, `appfwk` and `daqconf`

gcrone:
PRs merged yesterday so closing now.

PR #248, opened 2025-10-07
SmartDaqApplication inheritance

gcrone:
Make `SmartDaqApplication` inherit from `DaqApplication` and cahnge signature of `generate_modules` method as discussed in issue #247


src/ReadoutApplication.cpp
```diff
@@ -168,7 +168,7 @@ ReadoutApplication::generate_modules(const confmodel::Session* session) const
   // Scan Detector 2 DAQ connections to extract sender, receiver and stream information
   //
 
-  std::vector<const confmodel::DaqModule*> modules;
+  std::vector<const conffwk::ConfigObject*> module_objects;
```
alessandrothea: Why do we need to switch from `std::vector<const confmodel::DaqModule*>` to  `std::vector<const conffwk::ConfigObject*>` for `module_objects`? 

We could use to `dunedaq::confmodel::DaqApplication::set_modules` to update the list of modules from a vector of `DaqModule*`, couldn't we?

https://dune-daq.github.io/docs/classdunedaq_1_1confmodel_1_1DaqApplication.html#ac952fb9edb8068c6c3cf00c56345c00e

src/ReadoutApplication.cpp
```diff
@@ -429,9 +432,12 @@ ReadoutApplication::generate_modules(const confmodel::Session* session) const
   frag_aggr.set_objs("inputs", { &fa_net_obj, &frag_queue_obj });
   frag_aggr.set_objs("outputs", fa_output_objs);
 
-  modules.push_back(obj_fac.get_dal<confmodel::DaqModule>(frag_aggr.UID()));
+  auto dal_obj = obj_fac.get_dal<confmodel::DaqModule>(frag_aggr.UID());
+  module_objects.push_back(&dal_obj->config_object());
 
-  return modules;
+  auto app_obj = obj_fac.get_dal<ReadoutApplication>(UID())->config_object();
+  app_obj.set_objs("modules", module_objects);
```
alessandrothea: I think this step could be avoided using `dunedaq::confmodel::DaqApplication::set_modules`

src/ReadoutApplication.cpp
```diff
@@ -429,9 +432,12 @@ ReadoutApplication::generate_modules(const confmodel::Session* session) const
...
+  configuration().update<ReadoutApplication>({UID()}, {}, {});
```
alessandrothea: It doesn't feel great to expose this call to "usercode". 
How about adding a helper method to `ConfigObjectFactory` to call `update<ReadoutApplication>({UID()}, {}, {});`?
(yes, that requires tweaking the object factory a little, but it's probably doable without much work...)

alessandrothea (commented):
LGTM, just few aestetic comments.

src/ReadoutApplication.cpp
gcrone: I seem to remember that at some point when I tried using the dal setters that they were not implemented. As this does not seem to be the case now, the code could be changed to call set_modules but this opens a whole can of worms relating to the `const` nature of the dal objects. It seems that in this code we are changing the config_object, not the ReadoutApplication so we get away with it.

src/ReadoutApplication.cpp
gcrone: Looks like adding a method like this to `ConfigObjectFactory` can be used for a cheeky way of avoiding the `const` issue (note the naughty `const_cast`)
```
  void
  update_application(const std::vector<const confmodel::DaqModule*>& modules) {
    auto app = m_config->get<SmartDaqApplication>(m_app_uid);
    const_cast<SmartDaqApplication*>(app)->set_modules(modules);
    m_config->update<SmartDaqApplication>({m_app_uid}, {}, {});
  }
```
or using the `ConfigObject` pointers and avoiding the `const_cast`
```
  void
  update_application(const std::vector<const conffwk::ConfigObject*>& modules) {
    auto app_obj = m_config->get<SmartDaqApplication>(m_app_uid)->config_object();
    app_obj.set_objs("modules", modules);
    m_config->update<SmartDaqApplication>({m_app_uid}, {}, {});
  }
```

src/ReadoutApplication.cpp
alessandrothea: Right. At the same time, it would seem preferrable to deal with dal objects when assembling the configuration in general, to profit form compile time checks.
Maybe we could use the same cheeky trick you're suggesting in the next comment ;-)

src/ConfigObjectFactory.hpp
```diff
@@ -75,6 +78,14 @@ class ConfigObjectFactory
   const T* get_dal(conffwk::ConfigObject& obj) const {
     return m_config->get<T>(obj);
   }
+
+  void
+  update_modules(const std::vector<const confmodel::DaqModule*>& modules) {
```
alessandrothea, line 83: Something to watch out for here: 

The `ConfigObjectFactory` constructor doesn't enforce the parent object to be a `SmartDaqApplication`.
Either we should enforce a type check (by changing the constructor signature) or have a `ConfigObjectFactory` specialization for ``SmartDaqApplication`, that implements the `update_modules` method.

gcrone: Good point! I don't see the `ConfigObjectFactory` being used by anything other than a `SmartDaqApplication` so I think changing the constructor to take a `const SmartDaqApplication*` argument is a reasonable solution.

alessandrothea (approved):
LGTM

alessandrothea:
Hi,

a couple of last minute suggestions:
- Would it be wise to add a check that the module vector is empty before setting the new values? Being the  `SmartDaqApplication` a `DaqApplication` the module list could be populated (by mistakte)
- Shall we maybe generalize the `generate_modules` to `generate` (in case we want to generate other bits in the future? :smile:

PR #253, opened 2025-10-21
Rename TRBConf queue timeouts and sequence length parameters. Add Fra…

eflumerf:
…gmentAggreagator fragment_send_timeout_ms, set in ReadoutApplication based on longest TRBConf trigger record timeout in session.

# Description

_If full decription and testing details are included on a parent issue, please link to that here._
See DUNE-DAQ/dfmodules#455 for details

## Type of change

## Testing checklist

- [x] Unit tests pass (e.g. `dbt-build --unittest`)
- [x] Minimal system quicktest passes (`pytest -s minimal_system_quick_test.py`)
- [x] Full set of integration tests pass (`daqsystemtest_integtest_bundle.sh`)

_Comments here on the testing_

## Further checks


bieryAtFnal:
I'm a little uncomfortable with the assignment of the `fragment_sent_timeout` to the trigger record building timeout.  I understand the reasoning, which Eric has mentioned in various meetings - after the TRB gives up, there is no reason to keep retrying a fragment send.

I think that part of my discomfort is that if a user specifies a value for the `fragment_send_timeout`, it seems to get overwritten without any feedback to the user.

Another part is that a fairly long TRB timeout (Marco has mentioned 10 seconds in a meeting) could lead to a non-trivial shutdown time in the `FragmentAggregator::do_stop()` method, right?  (When the callback is removed from the fragment_input connection)

mroda88 (approved):
With this PR we also enforce  that the value of the timeouts is set from the configuration. Which is what was also discussed on slack following Kurt comment. Hence I'm approving.

bieryAtFnal (approved):
LGTM, code changes look good and regression tests continue to succeed with these changes.

PR #259, opened 2025-11-12
Updated trigger candidate maker configuration schema

alessandrothea:
# Description

Updates the  configuration schema to support the new `RandomTriggerCandidate` configuration parameters.

Replaces the `tc_readout` `RandomTCMakerConf` field with explicit parameters:
- `candidate_type_name`
- `candidate_offset_ts`
- `candidate_window_before_ts`
- `candidate_window_after_ts`

Addresses issue [#415](https://github.com/DUNE-DAQ/trigger/issues/415)

## Type of change

- [ ] Documentation (non-breaking change that adds or improves the documentation)
- [ ] New feature or enhancement (non-breaking change which adds functionality)
- [ ] Optimization (non-breaking change that improves code/performance)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Breaking change (whatever its nature)

## Testing checklist

- [x] Unit tests pass (e.g. `dbt-build --unittest`)
- [x] Minimal system quicktest passes (`pytest -s minimal_system_quick_test.py`)
- [x] Full set of integration tests pass (`daqsystemtest_integtest_bundle.sh`)
- [ ] Python tests pass if applicable (e.g. `python -m pytest`)
- [ ] Pre-commit hooks run successfully if applicable (e.g. `pre-commit run --all-files`)

_Comments here on the testing_

## Further checks

- [ ] Code is commented where needed, particularly in hard-to-understand areas
- [ ] Code style is correct (`dbt-build --lint`, and/or see https://dune-daq-sw.readthedocs.io/en/latest/packages/styleguide/)
- [ ] If applicable, new tests have been added or an issue has been opened to tackle that in the future.
  (Indicate issue here: # (issue))

PR #266, opened 2025-11-25
Revert "Updated trigger candidate maker configuration schema"

wesketchum:
Reverts DUNE-DAQ/appmodel#259

Reverting due to failing integration tests, and need for additional updates on configurations there, in other integtests, and perhaps also dunedaq_conf_reader


bieryAtFnal (approved):
I have run regression tests from `daqsystemtest, dfmodules, trigger, asiolibs`, and `crtmodules` with these changes and the corresponding changes in `daqsystemtest` and `trigger`, and the tests that were expected to pass did so.  (My software area started with the v5.5.0 RC2 build, included the repos that have changes for RC3, and included these changes.)

PR #271, opened 2026-01-30
Integrate CIB into dunedaq

nbarros:
# Integrate CIB DAQ modules into the latest dunedaq model

This PR introduces a new application that is meant to interface the CIB with dunedaq. The schema and corresponding application has been based off the CTB implementation, taking in consideration the differences between the two systems. 

The schema describes the configuration of both the accompanying [cibmodules](https://github.com/nbarros/cibmodules), and the board itself (only the portion related to the dunedaq - the remainder of the operation is controlled through slow control). 

_The new code requires the at NP02 to be operating, either in __simulation__ mode (generating fake triggers), or in actual data taking mode. The IoLS group should be contacted if there are intentions to test the code, so that arrangements can be made._

## Type of change

- [x] Optimization (non-breaking change that improves code/performance)

## Testing checklist

- [ ] Unit tests pass (e.g. `dbt-build --unittest`)
- [ ] Full set of integration tests pass (`daqsystemtest_integtest_bundle.sh`)

_The changes affect exclusively the interface with the CIB, and operational tests have been carried out successfully_

## Further checks

- [X] Code is commented where needed, particularly in hard-to-understand areas
- [X] Code style is correct (`dbt-build --lint`, and/or see https://dune-daq-sw.readthedocs.io/en/latest/packages/styleguide/)


mroda88 (commented):
The code seems ok to me. But I noticed that the link handler of the HSI signal are not receiving request. I suspect it's a configuration issue, i.e. a source id not set. But Before approving I would like to check.

mroda88 (approved):
With the changes to the MLT, the code is now good to go on my side. We see data requests coming and they are found so the timestamps are ok too.

PR #287, opened 2026-04-22
Change TPG threshold value type from u16 to s16.

aeoranday:
# Description
The value type in configuration for TPG thresholds were incorrectly set to unsigned integer 16 instead of signed. The problem was notice at-scale with ProtoDUNE VD (NP02) when setting the permissible configuration maximum for `u16`. Manually setting the value to `s16` maximum behaved appropriately and as expected for TPG.

Configurations should generally not be using threshold values near the maximum, so it is not expected to break existing configurations.

## Type of change

- [ ] New feature or enhancement (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)

## Testing checklist

- [ ] Unit tests pass (e.g. `dbt-build --unittest`)

Check that trying to set a threshold value greater than `32767` is not permitted. There is otherwise no TPG or Readout functionality change. This is purely a first-level configuration change.

## Further checks

### DIFF
--- a/schema/appmodel/trigger.schema.xml
+++ b/schema/appmodel/trigger.schema.xml
@@ -120,9 +120,9 @@
 
  <class name="AVXThresholdProcessor" description="TPG threshold signal processor. Filters the input signal by passing values that are above threshold and suppresing otherwise.">
   <superclass name="ProcessingStep"/>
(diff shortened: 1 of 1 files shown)


MRiganSUSX (approved):
tested max limit at ehn1 and worked as expected. 
(whether this should be a signed int is beyond me)

PR #290, opened 2026-06-15
Feat: TPwriter's enable/disable logic now determined by readout apps

henry-wallace-phys:
# Description
Previously TPStreamWriterApplication required additional external logic to determine its enable disable state. 
* "Are any ReadoutApps enabled?"
* "For each ReadoutApp do any of them have tp generation enabled"?

If there are no readout apps + none of tp generation enabled then the TPStreamWriter is itself disabled.
---
This is now moved internally using the 'tpset-net-rule' to find all connected readout applications. We find all configuration objects that use this rule, loop over them + find all ReadoutApplications used by them. 

## Type of change

## Testing checklist

- [x] Unit tests pass (e.g. `dbt-build --unittest`)

_Comments here on the testing_

## Further checks


src/TPWriterApplication.cpp
```diff
@@ -79,5 +80,40 @@ TPStreamWriterApplication::generate_modules(std::shared_ptr<appmodel::Configurat
...
+  // First we can just check if the application itself is disabled
+  if (!holder.disabled_components().is_enabled(this)){
+    return true;
+  }
+
+  // Now for the tricky bit, we need to loop over the connections
+
+  for(auto& rule : get_network_rules()){
+    /// HACK (minor): We assume TPs will always contain this exact rule
+    if(rule->UID()!="tpset-net-rule"){continue;}
```
gcrone: Why not use the same logic as at line 50 where you will at least be consistent

src/TPWriterApplication.cpp
```diff
@@ -79,5 +80,40 @@ TPStreamWriterApplication::generate_modules(std::shared_ptr<appmodel::Configurat
...
+
+    // We now loop over the parents
+    for(auto parent : configuration().referenced_by(*rule)){
+
+      // Safer than blindly casting to ReadoutApplication (RA)
+      if(!parent->castable("ReadoutApplication")) continue;
```
gcrone: I think casting and checking for nullptr is just as safe. The continue should really be on a line of its own, surrounded by braces. If statements without braces are against the coding conventions and really screw up my reading of code.

src/TPWriterApplication.cpp
henry-wallace-phys: Ah because I was being silly! Will fix in a moment

mroda88 (changes requested):
Nope, the function to extend is `compute_disabled_state`. To make sure, it is the right one, add the the qualifier `override`.

